PR #36, opened 2017-06-21
reduce memory request by looping with while

abonaca:


bd-j:
I think this breaks the iterator interface to the sampler.  Ideally this would use `xrange(sys.maxsize)` for python2 and `range(sys.maxsize)` for python3

abonaca:
I managed to run the demo successfully with this edit, but of course, there might be more issues. Also, according to [this stack overflow](https://stackoverflow.com/questions/36956983/python-memory-error-while-iterating-to-a-big-range), `xrange` wouldn't work, but I haven't tested that.

joshspeagle:
Ff you managed to run this with a `while`, maybe they changed things in Python 3? Otherwise, I thought that using Python 3's `range` would do the trick (at least, I didn't discover any problems, but I should check more carefully).

abonaca:
I'm using Python 2, and I think the `range` being imported now is not from the `__future__`. Perhaps that would be the most elegant solution?

bd-j:
yeah, the py3 range should work (though maybe with sys.maxint instead of sys.maxsize?)

but it's interesting that the while loop works in cell 9 of the demo

joshspeagle:
Yes. There are two problems here. The first is the Python 3 `range` from `__builtin__` is not the same as from `builtins`, which is part of `future` (via `pip install future`), so it's a standard list-generating function. The second is that actually iterators appear to work with `while` loops!

abonaca:
Why is a working while loop a problem?

joshspeagle:
@bd-j I'm assuming this hasn't always been the case?
@abonaca I just didn't realize `while` loops were compatible.

bd-j:
@abonaca it's not a problem, it's just that I thought we'd tested that while loops don't yield properly for an iterator.  If that's not true then great.

joshspeagle:
It looks like it works with no problem. However, given that `range` is used elsewhere in the code and should  properly be an iterator, I think I should instead just force everyone to install the `future` package to ensure that there are no issues.

abonaca:
Yeah, that would probably be the best solution, however, the previous import of `range` wasn't working. I'm trying to figure out the right syntax, as I actually have `future` installed.

joshspeagle:
Really? `builtins` is part of `future`, so that's strange.

abonaca:
Perhaps I was using the outdated version of `future`..

joshspeagle:
Maybe? I just had some students download in a few days ago for a similar compatibility reason and it seemed to work for them.

PR #57, opened 2017-10-26
add merge_runs fix

jrleja:
Fix for an edge case when using utils.merge_runs(). If you pass a list of results which has only a single baseline run in it, utils.merge_runs() dies. This will only happen when repeatedly merging runs after unraveling.


joshspeagle:
Looks good! Thanks for catching this.

PR #59, opened 2017-11-02
dynamicsampler : print progress to stderr (for notebooks) or stdout (for scripts)

bacook17:
Add key-word arguments to dynamic sampler methods to allow progress message to be printed either to stderr (with carriage return separation, useful for notebooks) or to stdout (with newline separation, useful for scripts).

Future work: could allow custom function to be passed for progress logging.

### DIFF
--- a/dynesty/dynamicsampler.py
+++ b/dynesty/dynamicsampler.py
@@ -1332,7 +1332,8 @@ class DynamicSampler(object):
                    maxiter_batch=None, maxcall_batch=None,
                    maxiter=None, maxcall=None, maxbatch=None,
                    stop_function=None, stop_kwargs=None, use_stop=True,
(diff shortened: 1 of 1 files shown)


joshspeagle:
Technically, you can do this yourself by running the sampler as a generator and using your own print function to print to stdout as you wish. However, I agree that given (1) the docs don't have a good example of this at the moment and (2) that seems like a lot of work to just dump progress to `stdout` instead of `stderr` (without the overwrite), this change is probably reasonable. I'll probably rewrite the internals a little bit to do this in a cleaner way rather than use the fix provided here (since it should apply to both `NestedSampler` and `DynamicNestedSampler`). I'll keep this pull request open until I add this feature in when I finally get around to the next series of updates.

PR #66, opened 2017-12-17
bug fix in sample_slice

jtmendel:
Changed variable uhat to u_hat to avoid NameError during the RuntimeError call.


joshspeagle:
Great -- thanks for the fix!

PR #80, opened 2018-02-19
Replace repeated code with call to add_batch

bacook17:
The DynamicSampler's run_nested function has code (inside the "Add points in batches" loop) which replicates the code in DynamicSampler.add_batch

This update replaces that repeated code with a single call to add_batch, which prevents having to change code in two places for any future updates to add_batch


bacook17:
There was a bug introduced by using add_batch: add_batch didn't have "stop_val" defined. Updated code to pass that value from inside run_nested.

joshspeagle:
Nice catch with the `stop_val` bug. I'll merge this change in since it's a straightforward simplification of the code.

PR #105, opened 2018-10-23
added save_bounds option to dynamic sampler

bacook17:
Dynamic sampler doesn't allow "save_bounds" option to be passed during initial batch. When combined with "update_interval" of 1, this can result in huge memory usage from saving each bounding object, which is then copied upon each call to "sampler.results"


bacook17:
Also added option for "maxcall_per_iter": stop any sampling if more than some number of calls are required for a given iteration

bacook17:
Not saving bounds caused new batch creation to (silently) fail.

PR #106, opened 2018-11-22
Fixed the TypeError from "numpy/core/fromnumeric.py"

tdaylan:
Hey Josh,

I just fixed the TypeError from "numpy/core/fromnumeric.py", line 83, in _wrapreduction

TypeError: 'float' object cannot be interpreted as an index

by changing np.min() -> min() and np.max() -> max() in plotting.py.

Best,
Tansu


joshspeagle:
Fantastic -- thanks.

PR #107, opened 2018-12-07
Dynamic sampler now updates live points 

bacook17:
Dynamic sampler previously did not update live_u, live_v, live_logl after the internal sampling began. Now, it assigns those variables to the same arrays as the internal sampler.

### DIFF
--- a/dynesty/__init__.py
+++ b/dynesty/__init__.py
@@ -10,4 +10,4 @@ from . import sampling
 from . import utils
 
 
-__version__ = "0.9.2b"
+__version__ = "0.9.3"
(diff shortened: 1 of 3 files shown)


joshspeagle:
this mostly looks fine minus the changes to plotting, which confuse me a bit. I'll try to either merge or separately implement these changes soon otherwise.

bacook17:
Hey sorry, I forgot I made those plotting changes in the same branch.
The basic thing I wanted to update with the plotting changes was to allow a different smoothing scale for the cornerplot contours (0.02 or 0.03 usually look pretty good) than for the density grid that's plotted underneath them. The density grid by default ends up being way too fine and really big in memory (when I save my 9x9 cornerplots as PDFs, they wind up 10+ MB)

joshspeagle:
I think this is enough of an edge case that I'll avoid merging this into `master` for now. I might pull features in the future if more people pull apart `dynesty` to the extent that you have though.

PR #109, opened 2019-02-05
initial version adding mlfriends

JohannesBuchner:
Here is an initial attempt merging my code into dynesty.

MLFriends is described in my latest paper Buchner (2019). It is RadFriends, but with first identifying the metric of the space. By default, it measures the standard deviation in each axis from the live points. This alone gives substantial speed-ups in terms of number of likelihood evaluations. 

So far this is all pure Python. Next I want to have a look at speeding up with numba. To my surprise, it is already running quite quickly though. 

Let me know what you think.

### DIFF
--- a/dynesty/dynamicsampler.py
+++ b/dynesty/dynamicsampler.py
@@ -38,7 +38,7 @@ __all__ = ["DynamicSampler", "weight_function", "stopping_function",
 _SAMPLERS = {'none': UnitCubeSampler,
              'single': SingleEllipsoidSampler,
              'multi': MultiEllipsoidSampler,
-             'balls': RadFriendsSampler,
+             'balls': MLFriendsSampler,
              'cubes': SupFriendsSampler}
 _SAMPLING = {'unif': sample_unif,
              'rwalk': sample_rwalk,
--- a/dynesty/dynesty.py
+++ b/dynesty/dynesty.py
@@ -26,7 +26,7 @@ __all__ = ["NestedSampler", "DynamicNestedSampler", "_function_wrapper"]
 _SAMPLERS = {'none': UnitCubeSampler,
              'single': SingleEllipsoidSampler,
              'multi': MultiEllipsoidSampler,
-             'balls': RadFriendsSampler,
+             'balls': MLFriendsSampler,
              'cubes': SupFriendsSampler}
 _SAMPLING = {'unif': sample_unif,
              'rwalk': sample_rwalk,
--- a/dynesty/nestedsamplers.py
+++ b/dynesty/nestedsamplers.py
@@ -22,6 +22,9 @@ Includes:
         Uses an N-cube of fixed length centered on each
         live point to bound the set of live points.
 
+    MLFriendsSampler:
+        Same as RadFriends, but learns the relative length of the space axes.
+
 """
 
 from __future__ import (print_function, division)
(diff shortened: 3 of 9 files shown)


joshspeagle:
This looks very reasonable. I'll see whether merging this or pulling out some components to implement separately makes more sense soon. Definitely the basic approach of just changing all the spheres (cubes) to be ellipsoids (cuboids) is straightforward enough to implement.

JohannesBuchner:
You will also see that I have code (commented out) that does the iterative Jarvis-Patrick clustering that polychord does to find clusters, overlays them to find the current metric/covariance. Do you already implement this whitening elsewhere? 

My code uses the metric from the previous iteration in the next iteration, so clustering can improve. That's kind of neat.

The Jarvis-Patrick clustering should be sped up by numba or similar; I have [c code](https://github.com/JohannesBuchner/UltraNest/blob/master/nested_sampling/clustering/cclustering.c) for that which is reasonably fast.

One issue this PR has is that the rstate's are not propagated. I use python iterators for generating new points, because this allows me to vectorise efficiently but still spit out one point at a time. Probably can be improved.

joshspeagle:
Sorry about the delay getting back to you on this. Now that I've had some more time to work on `dynesty` I've been able to spend more time looking into this. The commit looks fairly complete (and extremely impressive), but there's enough details that change the internals of how `dynesty` operates that I think it'd be preferable for me to either:
- just to point users to your excellent implementation (which I'm assuming is [here?](https://github.com/JohannesBuchner/UltraNest)) if they are interested in `MLFriends`,
- try and refactor this commit substantially to incorporate some of these changes natively into `'balls'` and/or `'cubes'` (since users should likely be getting these adaptive metrics and improved sampling strategies by default rather than my less-than-ideal implementation I built mostly for testing).

What are your thoughts?

JohannesBuchner:
We can probably refactor this a bit. I just don't like how the current balls implementation is slow (and I think the cubes method is unnecessary) and needs to build k-d trees. I have similar code elsewhere (UltraNest & massivedatans repos), but it is a bit always slightly different. 

Firstly, maybe you can integrate the jarvis_patrick_clustering code somewhere near your other clustering methods, perhaps even allowing it as an alternative?

Then I can at least squash the remainder into a single file, mlfriends.py, reformat to your coding style and removing some unnecessary code.

Integrating it with dynamic nested sampling would be cool, if you have the time to have a go at refactoring it a little bit to suit your preferences. If you have absolutely no time for that, an alternative could be moving it into a contrib/ contributed code directory.

JohannesBuchner:
PS: I see that I commented out the calls to jarvis_patrick_clustering* IIRC for performance reason of the pure Python version.

joshspeagle:
That sounds good. I'm going to close this pull request and open two issues for each of those, respectively, to implement:
- covariance structure in radfriends and supfriends (I think a Mahalanobis-version is actually quite doable now that I'm seeing the huge speedups involved in exploiting `numpy` array operations properly).
- Jarvis-Patrick clustering as a possible alternative.

I'll keep you updated on progress. I'm optimistic I'll be able to include this in the next release (0.9.4), along with the updated gradient support, which I'm trying to pull together now.

JohannesBuchner:
The only reason I did not prefer Mahalanobis in the end is that I like the force_shrink option, i.e., the scaling radius should only shrink between iterations. With balls this is easy, but if each axis is allowed to change its scale, it becomes more difficult. 

My approach was that I compute the std of each axis, normalise that by the maximum, take the log and round to integers. That gives me an idea of the relative sizes of the parameters. If it stays the same, I do not allow the radius to increase. If it changes, I allow the radius to increase.

Expanding this euclidean approach to a full covariance version has too many elements to fill, so force_shrink probably wouldn't work. force_shrink isn't required by any means, but it does speed things up, so it's a nice-to-have.

PR #110, opened 2019-02-05
Terminalout

JohannesBuchner:
Here is an experiment to avoid scrambled terminal outputs when the terminal is not large enough.

https://stackoverflow.com/questions/566746/how-to-get-linux-console-window-width-in-python

On python2.7, one may want to install https://pypi.python.org/pypi/backports.shutil_get_terminal_size, but this code should gracefully continue if it is not available.

### DIFF
--- a/dynesty/results.py
+++ b/dynesty/results.py
@@ -12,6 +12,7 @@ import sys
 import warnings
 import math
 import numpy as np
+import shutil
 
 __all__ = ["Results", "print_fn"]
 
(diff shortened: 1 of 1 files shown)


joshspeagle:
This looks like a nice quality-of-life improvement. I'll try to merge this once I get a chance.

bek0s:
After upgrading to the latest dynesty version which includes this update, I get nothing at PyCharm's output tab. Debian Testing + Python 3.7 + PyCharm 2018.3 community edition

joshspeagle:
Okay, looks like this request may have introduced some weird behavior. I'll try to see if I can locate the problem, but otherwise the `pip install` version should be clean.

JohannesBuchner:
@bek0s, what does shutil.get_terminal_size(fallback=(80, 25)) return for you?

bek0s:
Hi @JohannesBuchner,

If I print the output of the above command I get `os.terminal_size(columns=80, lines=25)`, regardless which version of dynesty I am using.

Thanks

Issue #121, opened 2019-03-29
MLFriends


JohannesBuchner:
Hi Josh,

I wrote a new implementation of MLFriends (learning the scales of the space), and tried to adhere to dynesty coding style. I believe the solution is quite compact:

https://github.com/JohannesBuchner/dynesty/commit/5d6edb35c89bb7d7d887c72186ed8c34eb7de7ff

Already for a simple gaussian likelihood on a unit cube::

```
   scales = np.array([0.1, 0.001, 0.3])
   def loglike(x):
       return -0.5 * (((x - 0.5) / scales)**2).sum()
```

The speed-up from simply normalising each axis is already a factor of 4.

The efficiency improvement will be arbitrarily high when scales[1] is set to any lower value.

The reference for this variant of RadFriends is https://arxiv.org/abs/1707.04476 (just mentioned in a sentence).

In Radfriends it is also natural to define clusters of balls. Avoiding a scaling across ball clusters, and using intra-cluster scales further improves results in multimodal distributions.

The following likelihood:

```
  scales = np.array([0.1, 0.001, 0.3])
  mu1 = np.array([0.3, 0.2, 0.3])
  mu2 = np.array([0.3, 0.4, 0.3])
  def loglike(x):
      l1 = -0.5 * (((x - mu1) / scales)**2).sum()
      l2 = -0.5 * (((x - mu2) / scales)**2).sum()
      return np.logaddexp(l1, l2)
```

Then sees another factor of 4 increase in efficiency, and again arbitrarily higher with cluster size to cluster distance increases.

I believe including this would be useful for users of dynesty.

joshspeagle:
I'm always happy to add in alternate implementations. I don't think I'll have time to get to this before I submit/release the paper on arxiv and will be traveling for the next two weeks, but I'll try to add this in when I get the chance for users who are't satisfied with ellipsoids :).

joshspeagle:
Hey Johannes,

Looking through your implementation in detail I agree this is actually a really nice place to start and an elegant implementation. I think the use of mean-centering from clustering is a good approach for dealing with multi-modal distributions. The failure to account for different behavior between the modes (i.e. assuming they are similarly shaped) is inherent to having a global metric for the ensemble relative to using multiple ellipsoids of varying sizes and shapes, so there's not much you can do about it without fancier changes.

If it's alright with you, I'd like to incorporate this solution directly into `balls` and `cubes` following your outline with two main changes:
[...]

PR #123, opened 2019-03-30
New implementation of mlfriends

JohannesBuchner:
See issue #121 .

For a simple gaussian likelihood on a unit cube::
```
  scales = np.array([0.1, 0.001, 0.3])
  def loglike(x):
      return -0.5 * (((x - 0.5) / scales)**2).sum()
```

The speed-up from simple scaling is already a factor of 4.

The clustering that detects disjoint ball groups
helps scaling multimodal distributions.
The following likelihood:

```
  scales = np.array([0.1, 0.001, 0.3])
  mu1 = np.array([0.3, 0.2, 0.3])
  mu2 = np.array([0.3, 0.4, 0.3])
  def loglike(x):
      l1 = -0.5 * (((x - mu1) / scales)**2).sum()
      l2 = -0.5 * (((x - mu2) / scales)**2).sum()
      return np.logaddexp(l1, l2)

```
Then sees another factor of 4 increase in efficiency.

The efficiency improvement will be arbitrarily high
when scales[1] is set to lower values.

Reference is https://arxiv.org/abs/1707.04476

### DIFF
--- a/dynesty/bounding.py
+++ b/dynesty/bounding.py
@@ -32,13 +32,14 @@ import numpy as np
 from numpy import linalg
 from scipy import special
 from scipy import spatial
+from scipy import cluster
 from scipy import linalg as lalg
 from numpy import cov as mle_cov
 
 from .utils import unitcheck
 
 __all__ = ["UnitCube", "Ellipsoid", "MultiEllipsoid",
-           "RadFriends", "SupFriends",
+           "MLFriends", "RadFriends", "SupFriends",
            "vol_prefactor", "logvol_prefactor", "randsphere",
            "bounding_ellipsoid", "bounding_ellipsoids",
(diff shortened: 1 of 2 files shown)


joshspeagle:
Looks good! I'll try to merge this in when I have some time to implement this more thoroughly. Hopefully that'll be sooner rather than later.

JohannesBuchner:
Hi @joshspeagle, could you please merge this. I would like to build some things on top of dynesty+mlfriends, but it's cumbersome to release without this PR merged in some form.

joshspeagle:
Sorry for the delay. I was traveling the last two weeks and am currently occupied for Passover with my in-laws. I’ll do this first thing next week.

joshspeagle:
Merging this in now. Will try to rework this to work with covariances shortly (see #121)

PR #129, opened 2019-05-24
Map the transformed u value in rwalk, either periodically, or reflect

### LINKED ISSUE #128: Periodic and reflecting boundaries
The periodic boundaries introduced recently are great. For transient gravitational waves signals, the posteriors often are of the same size as the prior for parameters we don't measure well and this  really helps.

Something we noticed though is that the current implementation doesn't work so well when the posteriors are "larger" than the prior. E.g, if you have a phase parameter which has a posterior width > 2 * pi. 

Because dynesty doesn't update the position of the parameter, the parameters tend to wander until they hit the -0.5 and 1.5 boundaries (e.g. [here](https://github.com/joshspeagle/dynesty/blob/master/dynesty/utils.py#L29)) and then strange things happen. It would be a great improvement if there was some way to pass back into dynesty where they should be mapped to. That is, if dynesty proposes the point 1.1, but this should be mapped back to 0.1, then the `u` vector should be updated. This would prevent the issue.

Additionally, for parameters that are not periodic, dynesty currently struggles to sample near to the boundaries. This could be solved by implementing reflecting boundaries where 1.1 would be reflected to 0.9. From our perspective, any non-periodic parameter should be reflective by definition, but perhaps there are cases where this isn't true.

What are you thoughts on this? We plan to start trying to look into the internals and figure out how this could be implemented ourselves and perhaps create an MR. But, your input would be quite valuable to make sure we understand how best to do this.

GregoryAshton:
This is an initial suggestion for how to close #128. This lets the `u` wander outside the unit cube, then applies the logic to map it back in based on if the parameter is periodic or not. Periodic parameters get wrapped around the unit cube, 1.1 -> 0.1, while reflective (all non-periodic parameters) get reflected, e.g 1.1 -> 0.9. 

This is suggested as a "work in progress", I'd guess it needs implementing in all the other proposal methods so they have similar behaviours, but I thought I'd just put something initial in to see what you thought. This also makes some of the periodic checking code obsolete I think.

### DIFF
--- a/dynesty/sampling.py
+++ b/dynesty/sampling.py
@@ -195,6 +195,12 @@ def sample_rwalk(args):
             du = np.dot(axes, dr)
             u_prop = u + scale * du
 
+            # Wrap periodic parameters and reflect non-perioridc parameters
+            u_prop[~nonperiodic] = np.mod(u_prop[~nonperiodic], 1)
+            u_prop_ref = u_prop[nonperiodic]
+            u_prop[nonperiodic] = np.minimum(
+                np.maximum(u_prop_ref, abs(u_prop_ref)), 2 - u_prop_ref)
+
             # Check unit cube constraints.
             if unitcheck(u_prop, nonperiodic):
                 break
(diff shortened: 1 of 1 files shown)


joshspeagle:
This looks great! Merging this in. I'll also try and add this code snippet in to `rstagger`.

PR #131, opened 2019-05-28
Travis and Scipy fix

pcubillos:
Hey Josh,
Numpy updated their latest Scipy version to 1.3.0, which requires python >3.5.  For some reason, under the current .travis.yml, even for the Python2.7 test, travis ends up installing Numpy 1.3.0, which makes it fail.  I edited out a few things that seemed unnecessary for travis (the miniconda lines), which fixes the automated tests.

Then, there was another bug.  Since version 1.0.0, Scipy deprecated scipy.misc.logsumexp(), which was finally removed in version 1.3.0 (the function is located in scipy.special.logsumexp since 0.19.0).  This causes dynesty to break for the latest Scipy version.

If you don't really care about the miniconda setup in the travis file, this PR fixes both of these issues (and passes the tests in both Python 2.7 and 3.6).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,20 +7,10 @@ python:
     - 2.7
     - 3.6
 
-#before_install:
-#    - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
-#    - chmod +x miniconda.sh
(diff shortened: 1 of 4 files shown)


joshspeagle:
Great -- thanks! Merging this in.

PR #135, opened 2019-06-20
add number of effective posterior samples as a stopping condition

ColmTalbot:
Hi @joshspeagle, this PR adds an additional stopping condition that a certain number of posterior samples have been generated.

This is something we're looking at using in Bilby for gravitational-wave analysis where we want to get a decent number of samples in the shortest time possible.
It would be great if you think this is useful enough to merge into master.

This relies on the [Kish effective sample size](https://en.wikipedia.org/wiki/Effective_sample_size).

### DIFF
--- a/dynesty/sampler.py
+++ b/dynesty/sampler.py
@@ -485,7 +485,8 @@ class Sampler(object):
                              "list of samples!")
 
     def sample(self, maxiter=None, maxcall=None, dlogz=0.01,
-               logl_max=np.inf, save_bounds=True, save_samples=True):
+               logl_max=np.inf, save_bounds=True, save_samples=True,
+               n_effective=np.inf):
         """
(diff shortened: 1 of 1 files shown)


joshspeagle:
This looks pretty reasonable as a stopping condition. I’ll plan to merge this in and then try to rework things a bit so that this can hopefully be applied both to Static and Dynamic Nested Sampling runs.

joshspeagle:
Nothing major that should require any edits on your end. My plan was just to reorganize the structure slightly to better match the other stopping conditions and possible flags (like adding the final live points). Then, I just wanted to add this in for the dynamic nested sampler as an alternate choice over the KLD error currently used, since it appears to make sense in that context.

Hoping to get to this sometime later this week while I deal with the final `dynesty` referee report, which also requested code changes.

PR #138, opened 2019-06-26
Added initial live pts check.

pacargile:
Added a "while True" loop to make sure there is at least one not -inf initial live pt.


joshspeagle:
Sounds good. As we discussed, I'll also see about adding in a "I tried N times and couldn't get anything to work" error message.

PR #151, opened 2019-09-12
Removes the automatic reflective boundaries

GregoryAshton:
In testing using bilby, we found that reflective boundaries can result in odd
behaviour at the boundary. We believe this is because the detailed balance of
the MCMC steps are no longer satisfied (multiple ways exist to get to the same
point). This removes the automated reflective boundaries for
non-periodic parameters (but retains periodic boundaries).

### DIFF
--- a/dynesty/sampling.py
+++ b/dynesty/sampling.py
@@ -195,13 +195,9 @@ def sample_rwalk(args):
             du = np.dot(axes, dr)
             u_prop = u + scale * du
 
-            # Wrap periodic parameters and reflect non-perioridc parameters.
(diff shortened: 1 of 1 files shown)


GregoryAshton:
@joshspeagle this was added (from what I recall) by @ColmTalbot . We have discussed this and think the reflective boundaries are problematic. One could imagine having "reflective", "periodic" and "None" boundaries, but I though initially we could just do the simplest thing and remove the problematic case.

joshspeagle:
Okay, good to know. Happy to merge this in and revert back to the previous behavior.

PR #152, opened 2019-09-12
Adding support for tqdm for printing progress

dfm:
This pull request should remain backwards compatible, but it adds support for using [tqdm](https://github.com/tqdm/tqdm) for displaying progress if it's installed.

The benefit of this is that tqdm plays nice and doesn't swamp stderr. I did a simple test and this reduces the output from 1,330,315 characters written to stderr (over 6185 iterations) to 5,076 characters for a model that is fast to evaluate. I've found that in some simple test cases (especially in Jupyter notebooks) just the I/O of the progress bar can  dominate the computational load and this should help reduce that effect.

Feel free to ignore if its not something that you're interested in!

### DIFF
--- a/dynesty/dynamicsampler.py
+++ b/dynesty/dynamicsampler.py
@@ -17,6 +17,7 @@ from six.moves import range
 
 import sys
 import warnings
+from functools import partial
 import math
 import numpy as np
 import copy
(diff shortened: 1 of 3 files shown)


joshspeagle:
Definitely interested and this looks like a very nice quality-of-life improvement. Merging this in now. Thanks!

PR #161, opened 2019-11-15
Add an additional catch for cases where the logwt is close to infinity

GregoryAshton:
If the logwt is too large, warning messages are produced by the
logsumexp terms. This stops that from occuring.


joshspeagle:
Looks good to me. Merging this in.

PR #163, opened 2019-11-26
Add an autocorrelation time estimate to rwalk and rstagger

GregoryAshton:
- Adds two new tuning parameters to the rwalk and rstagger methods:
  maxmcmc and nact
- maxmcmc controls the maximum number of steps
- nact controls the number of autocorrelation steps to take before
  accepting a point
- the ACT estimate is based on a blob post from Dan Foreman-Mackay
- the walks parameter has the modified meaning of the "minimum" number
  of steps to take

### DIFF
--- a/dynesty/sampler.py
+++ b/dynesty/sampler.py
@@ -324,6 +324,7 @@ class Sampler(object):
         scales = [self.scale for i in range(self.queue_size)]
         ptforms = [self.prior_transform for i in range(self.queue_size)]
         logls = [self.loglikelihood for i in range(self.queue_size)]
+        self.kwargs["live_u"] = self.live_u
         kwargs = [self.kwargs for i in range(self.queue_size)]
         args = zip(point_queue, loglstars, axes_queue,
                    scales, ptforms, logls, kwargs)
--- a/dynesty/sampling.py
+++ b/dynesty/sampling.py
@@ -168,7 +168,8 @@ def sample_rwalk(args):
     nfail = 0
     act = np.inf
     u_list = [u]
-    v = prior_transform(u)
+    v_list = [prior_transform(u)]
+    logl_list = [loglikelihood(v_list[-1])]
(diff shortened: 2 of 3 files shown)


joshspeagle:
Awesome! I'm out on vacation for a few days, but will try to merge this in over the next week.

GregoryAshton:
@joshspeagle great. Enjoy the holiday.

I would strongly encourage you to run this on some test cases (I see it runs on a bunch of stuff as part of the CI). Only because this is "fiddling with the guts" which makes me nervous :)

joshspeagle:
Yes, I definitely plan to run a battery of additional tests on harder functions to just check this behaves well. :)

Assuming everything goes well, I'll plan to merge it in and homogenize it so that it can also be used with the other sampling methods. Thanks again for putting this all together!

GregoryAshton:
I pushed a few more minor changes. I noticed that when one has posterior support at the boundary, one often sees the posterior peak away from from the edge (even when the simulation is set up with the true value at the boundary). This issue was fixed by return a random sample from the chain rather than the last point of the chain. This insight came from chatting with @ColmTalbot about potential biases when the chain only breaks after accepting a point.

GregoryAshton:
Sorry I push a few more changes. I realised that the emcee method of calculating the ACT was not performing well across a range of problems. I discovered that CPNest implements a much simpler check (which attributes to Will Farr's module). I've implemented this with attribution, but again I'm not certain how to do this.

I also made a few other minor fixes.

joshspeagle:
So I've now made enough code changes trying to resolve other issues (due to my delay in merging this) that I might have to incorporate some of these changes by hand in the master branch and reject this MR. I'll plan to give you credit in the updated documentation after the final version (v1.0.5) goes live.

segasai:
This PR is stale for a bit too long, so I will close it.

PR #174, opened 2020-02-23
Minor updates to references to some (now published) papers

ejhigson:


joshspeagle:
Thanks for updating these Ed!

PR #188, opened 2020-05-21
Make the DynamicNestedSampler class work better with saving partial progress.

psheehan:
Added a few options (save_samples, resume)to the sample_initial function to help a pickled DynamicNestedSampler resume a run that was pickled while running the initial nested sampling.

PR #192, opened 2020-07-06
__getstate__ has wrong indentation for DynamicSampler

ahnitz:
The __getstate__ method of the DynamicSampler wasn't actually being called as it had the wrong indentation and so was being considered a method of the __init__ method. I noticed this when trying to pickle DynamicSampler in the same way we do for the fixed live point samplers here. 

In combination with #188, this makes it a bit easier to save and resume progress.


joshspeagle:
Thanks for catching this and introducing a quick fix. This has now been merged.

PR #197, opened 2020-07-22
travis: update numpy to avoid mismatch between numpy and scipy

MatthieuDartiailh:
I also added 3.7 and 3.8 to the matrix since I assumed they are supported and should hence be tested.


joshspeagle:
Awesome! Thanks so much for this.

PR #210, opened 2020-10-05
fixed repeated removing while pickling sampler in sampler.py

ruskin23:
It looks like deleting `rstate`, `pool`  and `pool.map` was repeated [here ](https://github.com/joshspeagle/dynesty/blob/265bbe4bf6573704cea9ab4640ebf9e18edcae7f/dynesty/sampler.py#L161) and [here](https://github.com/joshspeagle/dynesty/blob/265bbe4bf6573704cea9ab4640ebf9e18edcae7f/dynesty/dynamicsampler.py#L469). I have added a simple try except statement in sampler.py to overcome this.

### DIFF
--- a/dynesty/sampler.py
+++ b/dynesty/sampler.py
@@ -158,12 +158,17 @@ class Sampler(object):
 
         state = self.__dict__.copy()
 
-        del state['rstate']  # remove random module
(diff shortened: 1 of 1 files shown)


joshspeagle:
Thanks for catching this! I'll merge this in. 👍

PR #218, opened 2021-03-11
Specify clustering dimensions

ColmTalbot:
Hi @joshspeagle, this PR would add a new parameter `ncdim` to specify the number of parameters to do the clustering on. This is distinct from the `npdim` where the nuisance parameters are things that we don't expect to gain much information for, but we are still interested in being able to associate specific sample values with the measured parameters. Adding in the option to not cluster on these parameters makes ellipsoid construction faster/more reliable for the remaining parameters.

I've tested this with `ncdim` set and unset for all of the built-in sampling methods with the multi-ellipsoid sampler and it seems to work well. If you're interested in merging this, I'd be happy to do a quick test with the other samplers.

I've tried to follow the repository style, but let me know if there's anything you'd like to change.

### DIFF
--- a/dynesty/dynamicsampler.py
+++ b/dynesty/dynamicsampler.py
@@ -343,11 +343,12 @@ class DynamicSampler(object):
 
     def __init__(self, loglikelihood, prior_transform, npdim,
                  bound, method, update_interval, first_update, rstate,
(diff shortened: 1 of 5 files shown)


joshspeagle:
This looks fine to me! Merging it in.

PR #225, opened 2021-03-29
logvolume patch

segasai:
Hi, 

This is an incomplete patch that would move away from volumes to logvolumes (in continuation of previous PR)

If there is a desire to have that, I can finish that. 

 S

### DIFF
--- a/dynesty/bounding.py
+++ b/dynesty/bounding.py
@@ -151,11 +151,6 @@ class Ellipsoid(object):
         self.am = lalg.pinvh(cov)  # precision matrix (inverse of covariance)
         self.axes = lalg.cholesky(cov, lower=True)  # transformation axes
 
-        # Volume of ellipsoid is the volume of an n-sphere divided
-        # by the (determinant of the) Jacobian associated with the
-        # transformation, which by definition is the precision matrix.
-        detsign, detln = linalg.slogdet(self.am)
-        self.vol = np.exp(logvol_prefactor(self.n) - 0.5 * detln)
 
         # The eigenvalues (l) of `a` are (a^-2, b^-2, ...) where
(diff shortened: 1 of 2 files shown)


joshspeagle:
> If there is a desire to have that, I can finish that.

Yes, please! I would definitely appreciate moving all of the volume calculations into log-volume ones. Finishing this off would be super. Let me know when things look stable and I'll plan to try and merge it in.

segasai:
I think this is a version that passes the tests. But I think a code review is needed, to check that I correctly mapped everything.

I also kept all external facing variable names the same. 

Also I'm sure there could be some other pointless exp() in the code that can be removed with the new changes.

joshspeagle:
Okay, I've gone through and checked most of this. I don't spot any remaining errors and it passes the basic CI tests, so I'm going to merge this in. Thanks so much for all your work on improving the stability and performance of the bounding distributions!

segasai:
Thanks. It would be good to make a release at some point, incorporating all those fixes. 
Also on separate issue.
There are few things it would be nice to do: 
1) move from travis to github actions (as that's more responsive and doesn't have such problematic limits that travis introduced)
2) Move the test suite to pytest or something like this 
3) Add more tests/increase coverage fraction (actually I don't know if you have running coveralls ? ) 
Do you have any objections against any of that ?  I may submit a few patches for that if I have time.

joshspeagle:
Yes, there's definitely enough changes relative to the last release it's worth tackling a few of the small quality-of-life bugs that have piled up and cutting a new one, along with an associated set of version notes.

No objections at all towards shifting the CI suite and increasing the coverage fraction (I set it up when I was just learning about these things). Appreciate any/all PRs on those issues, in addition to all the ones you've submitted already!!!

PR #229, opened 2021-03-30
GitHub actions + testing + coverage

segasai:
Hi, 

here is the  PR that
* moves to github actions https://github.com/segasai/dynesty/actions/runs/701554416
* moves to pytest testing
* updates the testing a bit to make it more manageable, so one can easily add more tests
* moves the main folder into py/ subfolder ( that is a common practice and it helps to deal with code/coverage problems)
* enables coverage testing https://coveralls.io/jobs/77586941 (so finally it could be seen what code is actually exercised in the tests)


segasai:
I also added to the commit chain a test of pathologically elongated distribution that triggers the ellipsoid regularization code

joshspeagle:
Okay, great. Looks like there are some issues with the `bounding.py` file that I need to try to take a look at after the recent merges. I'll try to get to that tomorrow if you don't get to it first.

joshspeagle:
Yea I suspected as much. If it's too much trouble I should be able to resolve this during the workday tomorrow and merge this in. Then I can cut a new release and start looking into dealing with all the minor issues that have been documented on the way to the next incremental release.

segasai:
I tried to see if I could solve it on my end, and I don't think I can, so you probably need to do it.

segasai:
closing this one as I have made a rebased PR without conflicts

segasai:
I'll just point here at the milestone )) of testing coverage finally exceeding 80% while we started from below 69%. The next goal is 90%.

PR #235, opened 2021-04-07
Saving history of likelihood evaluations

segasai:
Hi,

following earlier discussion, 
I propose a test patch that implements optional saving of all likelihood evaluations (in an hdf5 file).
This is done by wrapping the likelihood function and occasionally dumping things into a file. 

If there are no objections, I can tie a few loose ends, i.e. ensure that the last few calls are not lost and adding some docs. 
 
(on a separate note, the use_pool dictionaries need a set of tests).

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -1679,6 +1679,7 @@ class DynamicSampler(object):
         finally:
             if pbar is not None:
                 pbar.close()
+            self.loglikelihood.history_save()
 
     def add_batch(self, nlive=500, wt_function=None, wt_kwargs=None,
                   maxiter=None, maxcall=None, logl_bounds=None,
@@ -1793,6 +1794,7 @@ class DynamicSampler(object):
             finally:
                 if pbar is not None:
                     pbar.close()
+                self.loglikelihood.history_save()
 
             # Combine batch with previous runs.
             self.combine_runs()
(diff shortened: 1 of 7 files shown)


segasai:
I've just tested this on the test eggbox problem with 100 live points. And the results were pretty surprising and instructive:
fig  this is a histogram of all likelihood evaluations. 
In retrospect it's not that surpising, but still I didn't expect that kind of behaviour of ellipsoids 'hugging individual modes and becoming extremely elongated. That is certainly a useful a diagnostics tool IMO.

joshspeagle:
> If there are no objections, I can tie a few loose ends, i.e. ensure that the last few calls are not lost and adding some docs.

Yes, that sounds good. If you're satisfied with what you have now, I can look into merging this in. The syntax looks reasonable to me.

> (on a separate note, the use_pool dictionaries need a set of tests).

Yes, I don't have any pool-based tests set up right now. Including those would be useful.

> In retrospect it's not that surprising, but still, I didn't expect that kind of behaviour of ellipsoids 'hugging individual modes and becoming extremely elongated. That is certainly a useful diagnostic tool IMO.

Agreed.

segasai:
I've updated the branch, added a bit of docs. assuming it passes the tests, it's ready to be reviewed/merged

joshspeagle:
Looks good to me. Merging this in.

PR #239, opened 2021-04-10
various ellipsoid speedups

segasai:
tensordot is not needed

### DIFF
--- a/py/dynesty/bounding.py
+++ b/py/dynesty/bounding.py
@@ -208,6 +208,14 @@ class Ellipsoid(object):
 
         return np.sqrt(np.dot(np.dot(d, self.am), d))
 
+    def distance_many(self, x):
+        """Compute the normalized distance to `x` from the center of the
+        ellipsoid."""
+
+        d = x - self.ctr[None, :]
+
+        return np.sqrt(np.einsum('ij,jk,ik->i', d, self.am, d))
+
     def contains(self, x):
         """Checks if ellipsoid contains `x`."""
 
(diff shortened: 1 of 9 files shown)


segasai:
This set of commits speeds up various ellipsoid things. The effective speedup of update() with bootstrap is ~ 20%. The test is below

```import dynesty.bounding as db
import time
import numpy as np

np.random.seed(1)

points = np.random.normal(size=(100, 10))
e0 = db.Ellipsoid(np.ones(10), np.eye(10))
e1 = db.MultiEllipsoid([e0])
t1 = time.time()
for i in range(1000):
    e1.update(points, bootstrap=10)
t2 = time.time()
print(t2 - t1)

```
I also bit the bullet and ran bounding through yapf, to avoid occasional whitespace changes in the future (when I edit the code is automatically yapf it)

joshspeagle:
Are these changes ~stable enough that I should start to look over this PR for merging in?

segasai:
Yes, It' stable. It also has fixes to the likelihood saving (which broke pool functionality in the main branch).

segasai:
Did you have a chance of looking at this to merge ? I don't want to start changing other things before this is merged, due to potential conflicts.

joshspeagle:
No, sorry -- it's been a crazy busy week on my end. It's on the top of my to-do list, so I plan to do it by the end of the night (Eastern time) at the latest. Sorry about the delay.

segasai:
Thanks!

PR #243, opened 2021-05-03
Index fix

segasai:
The previous implementation of add_batch did not rewind down to the very lowest point (it stopped at the second lowest)
The patch fixes that.

Demonstration
```
nbase=10;nblive=4
i=np.arange(1,nbase-nblive);r=-(nblive+i)+nbase;print(r)
[5 4 3 2 1]
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,4 @@
-[run]
-source=py/
+[paths]
+source =
+       py/dynesty/
+       **/site-packages/dynesty/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,14 @@
 name: Dynesty
 
 # Run this workflow every time a new commit pushed to your repository
-on: push
+on:
+  push:
+    branches:
+      - '*'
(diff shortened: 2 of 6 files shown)


joshspeagle:
oof -- good catch here. Let me know when this looks good and I should review it.

Thanks for all the contributions!

segasai:
It is ready to be reviewed as far as I can see. 

I've also made a change that ensured that CI is being run on PRs.

joshspeagle:
This all looks good, except I think for the Dynamic Nested Sampling test it's not actually changing the number of live points since the API for that is slightly different. If that's fine, I can just merge in now though.

segasai:
Yes, I did not necessarily intend to change the number of live points I just wanted to have a test where the dynamic part does considerable amount of work (rather than exit early). And it certainly happens in that test case.
But in general the tests may need to be restructured somewhat (in the future).

Also after the merge it'd be nice if you could also check the paths in the coveralls page, to ensure the files are properly shown, as this branch includes updates to that

Issue #244, opened 2021-05-05
dynamic sampler bug, incorrect logl interval


segasai:
While running dynesty  dynamic sampler thousands of times I'm occasionally hitting the error "cannot find live points 
in the required logl interval". To debug it I put additional print there. See below:

```
  File "ZZZ/pyenv38/lib/python3.6/site-packages/dynesty
/dynamicsampler.py", line 1669, in run_nested
    stop_val=stop_val)
  File "ZZZ/pyenv38/lib/python3.6/site-packages/dynesty
/dynamicsampler.py", line 1775, in add_batch
    save_bounds=save_bounds):
File "ZZZ/pyenv38/lib/python3.6/site-packages/dynesty/dynamicsampler.py", line 1118, in sample_batch
    raise RuntimeError('Could not find live points in the required logl interval %f %f %f %f %f %d %d %d'%(logl_min, live_logl.min(), live_logl.max(), base_logl.min(), base_logl.max(), uidx, r, i))
RuntimeError: Could not find live points in the required logl interval 2596792.346555 2596788.423530 2596790.531368 -50000000000000000000.000000 2596790.531368 435 -501 1
```
The important point is logl_min is > base_logl.max().
This tells me that the sample_batch here https://github.com/joshspeagle/dynesty/blob/1a3cce6f490f275919208e3bdd154139ab20b117/py/dynesty/dynamicsampler.py#L1780 is called with the logl low boundary  that is above the logl of the base run. 
I can't immediately figure out why this happens. (and  since the error occurs sporadically on a cluster can't really debug it).
So I'll put it here for the moment. I wonder if the weight function is incorrect somehow since it provides the bounds...

segasai:
Okay, now I think I understand what's going on. 
The weight function called here 
https://github.com/joshspeagle/dynesty/blob/1a3cce6f490f275919208e3bdd154139ab20b117/py/dynesty/dynamicsampler.py#L1779
uses the results() data which is based on saved_logl (i.e. including multiple batch runs)
however sample_batch() relies on the base_logl when selecting live-points
Therefore it's entirely possible for the weight_function to return the interval that sits above the highest likelihood of the base run. 
I think the solution is probably either use the saved_* in sample_batch rather than the base run or somehow skip the intervals that sit above the base run (in some sense that means that the base run missed the tip of the posterior)

joshspeagle:
[...]

PR #245, opened 2021-05-06
tackling duplication in storing the information about the runs

segasai:
Following the issue #244 and required changes, I've decided to finally get rid of error-prone way of storing the information about the saved/based/new runs. Now there is a common object that stores various parameters. 
This makes a lot of logic simpler, getting rid of hundreds of lines of code.

Since the test suite is now in better shape, I think most of the bugs were caught by the tests, so I'm pretty confident about the patch. The patch is quite big, but most of the changes are pretty straightforward.

(since when I edit the code it automatically goes through yapf, there are some white-space changes as well). 

There are some stylistic improvements that can be further done, like reusing RunRecord from sampler in dynamicsampler.
It's also probably worth putting additional checks on the sizes of various arrays. i.e. assert that lengths of various arrays in run records are the same, but it's good enough IMO


segasai:
All the regression tests pass now, so it's ready for review

joshspeagle:
This is great! I've just finished going through and the internal refactoring is so much cleaner now, both for tracking things around and for possibly adding new quantities down the line. Happy to merge this in.

avivajpeyi:
Hi @joshspeagle, 

As this causes the way results are stored to change from the most up-to-date release on PyPI (v1.1), would it be possible to release a new version of dynesty on PyPI?

joshspeagle:
There are currently a few more active issues being resolved, but the plan is to push a new version once those wrap up in the near-term future.

Issue #212, opened 2020-11-08
Problems with the tilte of traceplot, cornerplot.


gapplef:
By default, `dynesty` will show 2-sigma credible interval in `traceplot`/`cornerplot` with dash lines, and it can be modified with parameter `quantiles`. If `show_titles` is true, the detail value will be shown in the title.
But the value shown in the title seems fixed to 2-sigma credible interval as shown in the following code, which means it would inconsistent with the interval lines shown in the figure if the user using a self-defined `quantiles` value, like 1-sigma. 

`traceplot`
https://github.com/joshspeagle/dynesty/blob/7a2630f5b797cb05ed827689c037009733612fef/dynesty/plotting.py#L692-L695

`cornerplot`
https://github.com/joshspeagle/dynesty/blob/7a2630f5b797cb05ed827689c037009733612fef/dynesty/plotting.py#L1256-L1259

joshspeagle:
Ooo, good catch. I should modify that so it defaults to 2-sigma, but if the user passes their own 3 quantiles it will use those instead. And there can be a separate argument to overwrite the defaults. Thanks for bringing this to my attention.

MJKirk:
@joshspeagle - I made a change that does everything needed to fix this issue. If you want I can make a pull request - the doc strings probably need improvement but otherwise I tried to do it nicely.

joshspeagle:
PR would be super!!! Totally fine if the docstring isn't perfect. Thank you so much!

PR #247, opened 2021-05-13
Add option to set title quantiles of corner/traceplot

MJKirk:
As also suggested in https://github.com/joshspeagle/dynesty/issues/212


joshspeagle:
This is a straightforward fix and the docstrings look good! Resolves #212.

PR #249, opened 2021-05-18
fix the problems with random seeds in tests.

segasai:
previously tests influenced each other.
Now each script has its own seed set up by pytest.fixture.


joshspeagle:
Should I plan to review and accept both this and #248 then?

segasai:
This patch is non-controversial, ready to be reviewed and committed. 
The #248 needs a careful separate look/review. I am not 100% certain in its correctness.

Issue #250, opened 2021-05-18
Slice sampling bug (sampling appears stuck error)


segasai:
While trying to use rslice sampling, even with many fixes I still hit the problem of 'Slice sampling appears to be stuck'

```
  /home/skoposov/science/ting_streams/distance_tools.py(402)sampler_dyn()
-> dsampler.run_nested(
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/dynamicsampler.py(1582)run_nested()
-> passback = self.add_batch(nlive=nlive_batch,
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/dynamicsampler.py(1702)add_batch()
-> for results in self.sample_batch(nlive_new=nlive,
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/dynamicsampler.py(1141)sample_batch()
-> for it, results in enumerate(
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/sampler.py(773)sample()
-> u, v, logl, nc = self._new_point(loglstar_new, logvol)
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/sampler.py(385)_new_point()
-> u, v, logl, nc, blob = self._get_point_value(loglstar)
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/sampler.py(369)_get_point_value()
-> self._fill_queue(loglstar)
> /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/sampler.py(358)_fill_queue()
-> self.queue = list(self.M(evolve_point, args))
  /home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/sampling.py(753)sample_rslice()
-> raise RuntimeError("Slice sampling appears to be "
```

The debugging tells me what happens is that 
the new loglikelihood level is selected https://github.com/joshspeagle/dynesty/blob/c22158cfb1a6bc740f7add2db9d238e272a9dbf6/py/dynesty/sampler.py#L763
Then the new point satistfying the L>Lmin value is selected here
https://github.com/joshspeagle/dynesty/blob/c22158cfb1a6bc740f7add2db9d238e272a9dbf6/py/dynesty/sampler.py#L773
[...]

PR #258, opened 2021-05-24
Slice sampling fix 

segasai:
I've needed an urgent fix to #250 since my codes were crashing regularly, so here is a possible fix. 
I'm not sure that's the best possible fix, but I think it does the job.
Basically it adds optional arguments for propose_*() 
And for the propose_live it accepts a subset of indices. It also avoids the repeated loglikelihood calls. 
There is a bit of a leakage of abstractions, as an if statement is needed to provide different arguments to propose functions depending on which sampler method is used, but maybe it can be refactored later.
The patch include some white space changed due to yapfing of previously un-yapfed nestedsamplers.py
The patch is based on the #257 commit branch.

### DIFF
--- a/py/dynesty/nestedsamplers.py
+++ b/py/dynesty/nestedsamplers.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-
 """
 Childen of :class:`dynesty.sampler` used to proposing new live points.
 Includes:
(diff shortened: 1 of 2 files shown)


joshspeagle:
I was working on something this earlier morning, but let me see if what you have is good enough for now. Will review it shortly.

segasai:
I'm happy with other ways of fixing it, I've just urgently needed to plug this hole. (there could be additonal bugs there)

joshspeagle:
Okay, I actually like your fix better than what I had implemented (mine was done internally without the use of args using the internal live point states). The ability to pass args or lists could be useful to build on in the future if we want to make more changes, plus it just makes everything more explicit (e.g., flagged by method) and avoids more duplication. I'm happy to merge this in once tests have completed.

segasai:
Okay, then I'm happy with the merging. I've added a bit more comments as well (no need to wait for the tests from those)

PR #259, opened 2021-05-24
Linting fixes of dynamicsampler.py

segasai:
Some pylint driven style improvements (remove unused variables, etc)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
           pip install coveralls
           pip install .
       - name: pylint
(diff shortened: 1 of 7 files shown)


segasai:
This should be ready to review/merge after the tests on the last commit are done, 
The main change here is that now pylint is being run in a more strict mode (that would have caught a previous oversight with saved_* stuff). That also forced me to change the constructors of the samplers ensuring that the super() call is happening at the beginning to avoid overriding.
Since the branch was started before the recent merge, I've had to cherry-pick a couple of commits here. The conflicts should be resolved in favour of the nestedsamplers.py from this branch.

PR #264, opened 2021-05-26
Possible improvement to the way we are dealing with degenerate ellipsoids

segasai:
I was thinking of changing the covariance matrix is 'improved' when it's invalid. 
As often the problem is that the ellipsoid is degenerate with eigen values being very small, a better fix should be keeping the eigen vectors and increasing the problematic eigen values to some level. 
This has the advantage of better preserving the shape of the ellipsoid in non-problematic dimensions.
The code is less pretty though.


joshspeagle:
> As often the problem is that the ellipsoid is degenerate with eigen values being very small, a better fix should be keeping the eigen vectors and increasing the problematic eigen values to some level.
> The code is less pretty though.

Agreed on both counts. The change also looks nice.

Issue #266, opened 2021-05-26
Further slice sampling issues 


segasai:
While continuing narrowing down rslice sampling issues leading to 
these kind of errors

```Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/skoposov/curwork/dynesty/tests/test_pathology.py", line 52, in test_pathology_dynamic
    sampler.run_nested(dlogz_init=1, print_progress=printing)
  File "/home/skoposov/curwork/dynesty/py/dynesty/dynamicsampler.py", line 1599, in run_nested
    passback = self.add_batch(nlive=nlive_batch,
  File "/home/skoposov/curwork/dynesty/py/dynesty/dynamicsampler.py", line 1719, in add_batch
    for results in self.sample_batch(nlive_new=nlive,
  File "/home/skoposov/curwork/dynesty/py/dynesty/dynamicsampler.py", line 1161, in sample_batch
    for it, results in enumerate(
  File "/home/skoposov/curwork/dynesty/py/dynesty/sampler.py", line 798, in sample
    u, v, logl, nc = self._new_point(loglstar_new, logvol)
  File "/home/skoposov/curwork/dynesty/py/dynesty/sampler.py", line 410, in _new_point
    u, v, logl, nc, blob = self._get_point_value(loglstar)
  File "/home/skoposov/curwork/dynesty/py/dynesty/sampler.py", line 394, in _get_point_value
    self._fill_queue(loglstar)
  File "/home/skoposov/curwork/dynesty/py/dynesty/sampler.py", line 383, in _fill_queue
    self.queue = list(self.M(evolve_point, args))
  File "/home/skoposov/curwork/dynesty/py/dynesty/sampling.py", line 759, in sample_rslice
    raise RuntimeError("Slice sampling appears to be "
RuntimeError: Slice sampling appears to be stuck! Some useful output quantities:
u: [0.5       0.4862176]
u_left: [0.49999999 0.48072734]
u_right: [0.50000001 0.48904739]
u_hat: [1.66894077e-08 8.32005324e-03]
loglstar: 17.956478101034662
axes: [[ 1.12058641e-03  0.00000000e+00]
 [-8.31235509e-02  3.54360519e+02]]
axlen: 938.2393864917427.
```

I have a few comments (no patch yet) 
I think this line is counter-productive
https://github.com/joshspeagle/dynesty/blob/a55832be419e4e019de7142f9c9651d0acc31326/py/dynesty/sampling.py#L752
[...]

PR #269, opened 2021-05-26
Fixes slice

segasai:
This is several commits.
One that revises the pointvol setting and only sets if there is one single point (as otherwise that can sometimes overinflate the ellipsoid)
I also remove the exception on slice sampling stuck (as discussed in #266)
Add a longer/more extended pathology test.


joshspeagle:
Looks good! Merging this in.

PR #271, opened 2021-05-27
refactor slice sampling

segasai:
The current slice samplng code still gave me the problems like this
```
RuntimeError: Slice sampler has failed to find a valid point. Some useful output quantities:
u: [0.43015122 0.26744562 0.9999996  0.69852869 0.51943458 0.99987344]
u_left: [0.43015122 0.26744562 0.9999996  0.69852869 0.51943458 0.99987344]
u_right: [0.43015122 0.26744562 0.9999996  0.69852869 0.51943458 0.99987344]
u_hat: [-2.77555756e-16 -5.55111512e-17  0.00000000e+00 -1.11022302e-16
 -1.11022302e-16  0.00000000e+00]
u_prop: [0.43015122 0.26744562 0.9999996  0.69852869 0.51943458 0.99987344]
loglstar: 549499.5271041454
logl_prop: 549499.5271040456
axis: [-5.43760729e-05 -1.47307934e-06  9.95810121e-08 -2.81106958e-06
 -9.69705233e-06  8.84660396e-09]
axlen: 5.5325149793425564e-05
s: 0.0.
```
It turns out that the reason is that during the slice sampling iterations due to floating point errors, it may happen that the original starting point is not more in the slice. 
The proposed patch fixes that and more:

refactor slice sampling by separating the single slice sampling step in a separate function
that allows sharing the code between _slice and _rslice samplers.
Also the slice sampling is now more numerically stable due to the fact that we are
working in the 'coordinate system' of u0 + x* direction, i.e. we are always strictly
guaranteed to have an initial point at 0 and are not affected by the precision loss

### DIFF
--- a/py/dynesty/dynesty.py
+++ b/py/dynesty/dynesty.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-
 """
 The top-level interface (defined natively upon initialization) that
 provides access to the two main sampler "super-classes" via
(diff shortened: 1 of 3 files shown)


segasai:
This is ready for review/merge. 
(And with those set of changes I was finally able to get through my entire set of 60k objects without crashes)

PR #275, opened 2021-05-28
Refactor of jitter_run/utils

segasai:
I've started to take a stab at optimizing/understanding/clarifiying jitter_run.

I replaced all the numpy 'int' ,'bool' types by int, bool.
I also separated the code that detects decrease in number of live-points. (I verified that it produces the same results). 

I think I'd like to vectorize/clarify the jitter_run code, as currently it's a bit of a maze to me. It'd be also good to have some kind of references explaining the maths there...

### DIFF
--- a/py/dynesty/utils.py
+++ b/py/dynesty/utils.py
@@ -511,6 +511,7 @@ def jitter_run(res, rstate=None, approx=False):
                                                        np.empty(nsamps),
                                                        np.empty(nsamps))
     for i in range(nsamps):
+        # TODO: explain maths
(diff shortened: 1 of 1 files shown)


joshspeagle:
As a quick reference if you'd like to pursue further vectorization, some of the math is outlined in Appendix A of the dynesty paper, in particular A.4 and A.5.

joshspeagle:
Okay, this looks good to me. Pulling out the "check if live points are decreasing" part of the code into its own separate internal function is also a good change. Thanks.

PR #278, opened 2021-06-08
fix the logzerr calculation, (make sure it's an array)

segasai:
a tiny fix to previous jitter changes discovered when running demo notebooks


joshspeagle:
whoops! good catch.

PR #280, opened 2021-06-08
remove  all the direct calls of np.random.*

segasai:
#218 incorrectly used np.random.directly instead of rstate

I've also added a test that will fail if it'll happen again.


joshspeagle:
Excellent fix to both the tests and catching this bug, which would prevent full reproducibility.

Issue #282, opened 2021-06-08
Hanging of the pathology test, volume scaling


segasai:
This 
```
env PYTHONPATH=py:tests:$PYTHONPATH python  -c 'import numpy.random as R,test_pathology as T;R.seed(5);T.test_pathology_dynamic()'
```
(executed in the main dynesty directory) 

will currently hang the test at this
```
7803it [00:02, 3350.77it/s, batch: 0 | bound: 17 | nc: 1 | ncall: 29661 | eff(%): 26.025 | loglstar:   -inf < 15.517 <    inf | logz:  2.672 +/-  0.138 | dlogz:
8125it [00:20, 3350.77it/s, batch: 1 | bound: 18 | nc: 161 | ncall: 29983 | eff(%): 27.099 | loglstar: 18.214 < 18.421 < 18.421 | logz:  2.868 +/-  0.434 | stop:  2.398]    
.....
```

A quick check reveals that this is related to scale_to_logvol leading to linear rescaling of  up to 1000 of the ellipsoid which then leads to the difficulties in finding the point in such ellipsoid.

I don't have a fix ATM. I think the pointvol logic is certainly questionable  (or unclear to me).
Alternatively some additional checks/logic can to be put in when ellipsoid is rescaled to many times the size of the unit cube.

joshspeagle:
I think we'd discussed this a bit before re: putting a ceiling on the maximum axes length based on the length of the diagonal of the unit cube \sqrt(n), although nothing's been put in quite yet.

segasai:
There is a limit now for the slice sampler (I've implemented it in one of the last patches), because there is a natural scale there. But this problem happens with the uniform (within ellipsoids) sampler. (and I still think that the pointvol assignment may need to be rethought as that's the root cause IMO, or alternatively the volume calculation of ellipsoids longer than the cube need to be updated)

joshspeagle:
Right. I think given how things are now, we can probably afford to put in two checks for this:
- A check on the maximum (log-)volume, which could be for something like an n-sphere with a radius of \sqrt(n)/2.
- A check on the maximum axis length, which could be, say, \sqrt(n).

This should help to counteract some of the issues that can arise from pointvol, which is still not ideal but at least a solution for the case where you end up with a bound centred on a single point. (If that edge case is successfully resolved elsewhere, I think we can safely remove it from all other locations in the code.)

segasai:
The #283 and #284 resolve this issue.

PR #283, opened 2021-06-09
Long ell

segasai:
This is the code that when inflating the ellipse prevents principal axes from being larger than the half of the cube diagonal. (It still tries to increase the volume to logvol, but by increasing other axes more). 
 That  solves the problem in #282  partially, but then moves the problem to the slice sampler. With this patch now the slice sampler hangs. 
I don't have time now to debug it further, but I'm guessing this is the issue with the volume in sample_batch()

### DIFF
--- a/py/dynesty/bounding.py
+++ b/py/dynesty/bounding.py
@@ -179,16 +179,38 @@ class Ellipsoid(object):
     def scale_to_logvol(self, logvol):
         """Scale ellipsoid to a target volume."""
 
-        f = np.exp((logvol - self.logvol) / self.n)  # linear factor
-        maxlen = np.sqrt(self.n)
-        fax = np.minimum(self.axlens * f, maxlen) / self.axlens
-        l, v = lalg.eigh(self.cov)
-        self.expand *= f
-        self.cov = v @ np.diag(l * fax) @ v.T
-        self.am = v @ np.diag(1 / l / fax) @ v.T
-        self.axlens *= fax
-        self.axes = lalg.cholesky(self.cov, lower=True)
-        # I don't quite know how to scale it
+        logf = (logvol - self.logvol)
+        # log of the maxium axis length of the ellipsoid
+        lmaxlen = np.log(np.sqrt(self.n) / 2)
+        laxlen = np.log(self.axlens)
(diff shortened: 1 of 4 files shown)


segasai:
I've set the volumes to zero in multiple places and updated how .expand is treated for bounds so it wouldn't be affected by the scale_to_logvol(). 
I also had to rewrite the bootstrap check in test_gau to avoid a false-positive fail of the test.

The test in a loop seem to work
 $ for a in `seq 0 36` ; do { echo $a ; env  DYNESTY_TEST_RANDOMSEED=$a PYTHONPATH=py:tests:$PYTHONPATH pytest -s tests/test_pathology.py   ; } ; done 

But I don't feel 100% confident that the setting of volumes to zero is completely harmless.

joshspeagle:
Great. As for setting the volumes to 0, maybe just picking a small value just above the numerical instability threshold could work?

segasai:
The problem is that I don't have a good understanding where and how the volume used and what it affects, so I dont' really know TBH. specifically I don't know if point volume is somethign that can enter in the calculations of logz, posterior etc

joshspeagle:
Ah, happy to confirm it doesn't! It's entirely used in constructing the bounds.

segasai:
Okay I've made the test PR #284 (based on this PR) 
that gets rid of point volumes completely. 
All the tests pass, but the test suite slows down and as far as I understand this is due to the removal of this
https://github.com/joshspeagle/dynesty/blob/cad81f485c7c3167260c03f7e8b2e165c21e7f6d/py/dynesty/bounding.py#L1566

Basically before the full hierarchical split was considered only if pointvol=0 or the requirement wrt pointvol was satisfied. I am not sure it made much sense, given that pointvol values were somewhat arbitrary, but that clearly prevented that codepath being taken often.

segasai:
thanks for merging 
I think if  #284 is not merged 
The half of thsi  e100d0727180ad04122237075287cbd7e29eb1b2
and this fe5d5ac9e861959e2799868a3f6221f88c46de2d commits need to be applied here as fixes

joshspeagle:
I'm in the process of merging that now, so that should be fine in a few minutes.

PR #284, opened 2021-06-11
Volume branch

segasai:
Getting rid of pointvolumes

### DIFF
--- a/py/dynesty/bounding.py
+++ b/py/dynesty/bounding.py
@@ -1512,6 +1512,14 @@ def _bounding_ellipsoids(points, ell, vol_dec=0.5, vol_check=2.):
                 points_k[0], ells[0], vol_dec=vol_dec, vol_check=vol_check) +
             _bounding_ellipsoids(
                 points_k[1], ells[1], vol_dec=vol_dec, vol_check=vol_check))
+    out = (_bounding_ellipsoids(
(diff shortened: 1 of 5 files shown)


segasai:
I think this is ready for review and merge. But I was also doing test, checking the behaviour of recovery of logz of arbitrary oriented elongated gaussian (with axis ratios of 1e-3:..:1) as a function of dimension.
And this is the behaviour of rslice with multi bounds
on xaxis is the ndim on the y-axis is the (logztrue - logzmeas)/logzerr 

image

And the interesting thing is that there is a clear trend in the logz bias vs ndim that was present before. 

And the volume change makes it somewhat worse. I'm suspecting that this must be connected to the ellispoidal splitting.

joshspeagle:
Does this trend get better if the number of slices is increased, by chance (assuming you're not already doing that)? The auto-correlation times should scale as ~d^1-2, so the number of slices also would need to increase to help encourage effective sampling. In the initial `slice` this was imposed implicitly by iterating through all parameters, but this is not true for `rslice`.

I also agree that this is likely connected with the ellipsoidal splitting giving a (more reliable but possibly worse?) approximation of the overall distribution.

joshspeagle:
Also, just going to ping #285 since that's likely the more appropriate area for continuing this discussion.

Issue #285, opened 2021-06-11
slice -> rslice + high-d bounding scaling issues


segasai:
Currently the default sampler in high dimensions is slice, which for a simple correlated 200d gaussian gives me ~ 0.06% efficiency vs 3.5% for rslice. 
I think rslice should be used instead.

joshspeagle:
Agreed.

segasai:
Continuing the discussion of #284

The this is the same plot as before, but now with single bound instead of multi 
![image](https://user-images.githubusercontent.com/109513/121953453-7bed5e00-cd55-11eb-901e-b0169118b376.png)

It's clear the issue disappears. So with the new change the splitting is likely a bit too agressive (it's still not 100% but very likely)

Below is the code used for the test
```
import numpy as np
from scipy import linalg
import scipy.stats
import dynesty  # noqa
from dynesty import plotting as dyplot  # noqa
from dynesty import utils as dyfunc  # noqa
"""
Run a series of basic tests to check whether anything huge is broken.

"""

nlive = 5000
printing = False

# GAUSSIAN TEST


def get_covar(rstate, ndim):
    eigval = 10**np.linspace(-3, 0, ndim)
    M = scipy.stats.ortho_group.rvs(dim=ndim, random_state=rstate)
    ret = M @ np.diag(eigval**2) @ M.T
    return ret


class Config:
    def __init__(self, rstate, ndim_gau):
        self.ndim_gau = ndim_gau
        self.mean_gau = np.linspace(-1, 1, ndim_gau)
        cov_gau = get_covar(rstate, ndim_gau)
        self.cov_gau = cov_gau
        self.cov_inv_gau = linalg.pinvh(cov_gau)  # precision matrix
        logdet = np.linalg.slogdet(cov_gau)[1]
        self.lnorm_gau = -0.5 * (np.log(2 * np.pi) * ndim_gau + logdet)
        self.prior_win = 1000  # +/- 10 on both sides
        self.logz_truth_gau = ndim_gau * (-np.log(2 * self.prior_win))
        assert (np.isfinite(self.lnorm_gau))
        assert (np.isfinite(self.logz_truth_gau))


class si:
    config = None


# 3-D correlated multivariate normal log-likelihood
def loglikelihood_gau(x):
    """Multivariate normal log-likelihood."""
    co = si.config
    x1 = x - co.mean_gau
    return -0.5 * x1 @ co.cov_inv_gau @ x1 + co.lnorm_gau


[...]

PR #286, opened 2021-06-14
remove volume_check & update vol_dec for hierarchical splits

segasai:
following #285
make the conditions stricter on hierarchical splits if the first split didn't work
also get rid of no-op vol_check option.

This partially recovers the losses of #284 (blue curve) as the deviations start to now happen at ndim=17
image

I'm not sure we'll be able to fully get it back, as the previous code used pointvol for this. In my mind there were cases where pointvol estimates were good and it was preventing excessive splitting but in some cases it was preventing splitting because of incorrect pointvols.

I also change the sampler to rslice in high dims.


segasai:
It looks like now the patch recovers the high dim behaviour that was there before. 
It also removes the vol_dec parameters so I think it's ready to be reviewed/merged.

PR #292, opened 2021-06-17
Random fixes

### LINKED ISSUE #291: sample_* don't respect rstate
While looking at the sampling code I noticed that the sample_rslice and other  methods do not use rstate at all. They just set rstate to np.random

https://github.com/joshspeagle/dynesty/blob/6843ca1e0df26410679cf39cb2ce43695782698f/py/dynesty/sampling.py#L150

This clearly is incorrect both for single and mult-threaded cases. The fix though is somewhat not trivial as one needs to somehow either pass random state or a seed generated from a parent random state. But one can't just send the same random state.
Also generating a random state is somewhat heavy ~ 0.1, so it'd be better to avoid recreating them all the time.

segasai:
Fix to not use np.random anywhere, instead rely on numpy generator. 
Also ensure that the map() methods get seeds to initialize the random generators, so things are fully deterministic even in the multithreaded scenario.
Fixes #291


joshspeagle (commented):
A comment on the `slice` behavior in the tests: that's likely due to the default axes not being a good match to the Gaussian, which I guess affects the Gibbs-style update more than any of the other ones due to the poor mixing from the proposals. Not too concerned with that edge case.

Issue #293, opened 2021-06-18
Arguments to NestedSampler DynamicNestedSampler


segasai:
One constant issue I'm hitting with dynesty is that 
NestedSampler and DynamicSampler do not check for keyword arguments.
I.e. I often tried to initialize DynamicSampler with nlive parameter which does not do anything. I also occasionally provided stop_kwargs to the nested sampler and that is basically silently ignored. 
That can be particularly painful with nlive parameter. 

I think my preferred solution is to not allow freeform kwargs for NestedSamplers so we'd get the exception. 
Also, it may be good to use nlive parameter of Dynamic nested sampler as nlive_init if they were not specified explicitely.

joshspeagle:
I was about to recommend exactly the type of solution you just submitted for the PR. Removing the freeform kwargs is a bit painful, but I agree avoiding the silent failures is probably worth it.

segasai:
Fixed in #295

PR #295, opened 2021-06-18
Kwargs fix

segasai:
Attempt to address #293 
1) I remove **kwargs from Nested and Dynamic samplers
2) add nlive parameter to Dynamic samplers that will be used as default nlive_init/nlive_batch if those are not specified.


joshspeagle:
Is this stable and ready to be reviewed and merged?

segasai:
Yes, It is.
The docs will need some adjustment (later though)

joshspeagle:
No problem wrt the docs. I can always add that in later when I'm trying to collect all these changes into `v1.2`.

Issue #289, opened 2021-06-16
High dimensions log(z) bias


segasai:
As seen in #285 there is a bias in logz at high dimension count. 

I've briefly checked and that seems to be not correlated with the number of live points (see below)
This is kind'a suspicious as you'd expect that any numerical/precision issues to be smaller when you increase nlive.
The bias is pretty substantial of 50 dex. 
I'm wondering if this is a manifestation of some kind of a bug somewhere as opposed to inaccuracies.

```
In [6]: rstate=np.random.RandomState(33);co=test_highdim.Config(rstate,32);X=test_highdim.do_gaussian(co,sample='rslice',bound='multi',nlive=500,rstate=rstate)

In [7]: co.logz_truth_gau
Out[7]: -243.22887870534663

In [8]: print (X)
(-195.13836523939548, 1.0181387163605187)

In [9]: rstate=np.random.RandomState(33);co=test_highdim.Config(rstate,32);X=test_highdim.do_gaussian(co,sample='rslice',bound='multi',nlive=1500,rstate=rstate)

In [10]: print (X)
(-197.90183809316025, 1.022940499442747)

In [11]: rstate=np.random.RandomState(33);co=test_highdim.Config(rstate,32);X=test_highdim.do_gaussian(co,sample='rslice',bound='multi',nlive=5000,rstate=rstate)

In [12]: print (X)
(-195.56476086603644, 1.0202571010884396)

```

segasai:
Apparently increasing the number slices to 20 helps is reducing the bias

```
In [2]: rstate=np.random.RandomState(33);co=test_highdim.Config(rstate,32);X=test_highdim.do_gaussian(co,sample='rslice',bound='multi',nlive=500,rstate=rstate)

In [3]: X
Out[3]: (-232.03421701937455, 1.0847506259228943)

In [4]: co.logz
Out[4]: -243.22887870534663
```

So I guess the bias is caused by correlated samples. I'm a bit surprised, as I'd thought slice sampling inside the ellipse should mix really well, but I guess in high-D spaces, that's not true anymore.

joshspeagle:
[...]

PR #297, opened 2021-06-18
add scaling with ndims for walks/slices

segasai:
address #289


segasai:
On this subject, I'm also proposing making rslice as a default sampler in high dimensions, rather than  using rwalk in 10..20 dimensions range.

joshspeagle:
>I'm also proposing making rslice as a default sampler in high dimensions, rather than using rwalk

Is that just because it gets it is more robust, or also that at this point its performance is actually better as well?

segasai:
I think rslice is quite robust now. It also requires less steps than rwalk. I think rwalk also occasionally produce the 'random walk is very inefficient' messages and is likely more sensitive to bad scaling. I have a strong opinion that rslice is now more stable, but I don't have hard numbers. (for my own stuff, I now almost always use rslice)

joshspeagle:
Each step for `rslice` is multiple function calls though, which is why I was curious what the overall sampling efficiency looked like by comparison. If it's not a huge efficiency hit in that dimension range while being more robust overall, I agree that the defaults can be changed.

segasai:
Here's the Gaussian test.
image

Unexpectedly (to me) rwalk indeed outperforms rslice (roughty factor of 3 in number of calls). So I temporarily  retract the 
 change of the default sampler for 10-20 ndims.
I'm still wondering if there is something not quite right with scaling up or down the step size for slice sampling (as that seemed to misbehave occasionally)

joshspeagle:
That adaptive step was supposed to make sampling more efficient overall based on what I had seen in the literature and from other tests (like with `rwalk`), but if it looks like it hurts more than it helps I'm happy to disable it by default.

segasai:
I don't have a hard proof (ATM), so lets not change anything without it, but lets keep it in mind.

joshspeagle:
I tried to merge this into the main branch, but I think I slightly screwed up the line editing. Should be a quick fix on your end whenever you have a moment; otherwise, I can get to this Monday.

segasai:
Fixed it. 
The problem is that the live_points argument for some reason only is accepted by the nestedsampler not dynamic one. (and previously the test just ran the test ignoring the live_points arg, but after #295 the code started to complain)

Issue #302, opened 2021-07-31
Cannot compute a bounding ellipsoid to a single point if `pointvol` is not specified


doublestrong:
https://github.com/joshspeagle/dynesty/blob/243025cbfc5f1941e9dc1b00348c54f8b877bd6c/py/dynesty/bounding.py#L1316

Hi, I encountered this error when using DynamicNestedSampler with hslice and my own function of grad_u. This error shows up immediately when the sampler reaches the dlog_z_init, which is 50.0 in my setting. I think hslice works as my grad_u function was called without errors. ndim of the problem is around 90.

Error message:
```ruby
create run dir: /home/chad/codebase/project/example/xx/_/res/res_100_/seed0/case0/dyn28
  using dynamic nested sampling
iter: 2681 | batch: 0 | bound: 1 | nc: 1 | ncall: 865688 | eff(%):  0.310 | loglstar:   -inf < -26.582 <    inf | logz: -36.999 +/-  0.914 | dlogz:  0.617 > 50.000                                   Traceback (most recent call last):
  File "/home/chad/codebase/project/example/xx/_/dynamic_ns_incremental.py", line 15, in <module>
    dynesty_run_batch(500, case_dir, data_file, data_format, parallel_config = {'queue_size': 64},
  File "/home/chad/codebase/project/src/xx/RunBatch.py", line 289, in dynesty_run_batch
    sample_arr = solver.sample(live_points=live_points, pool=pool, queue_size=parallel_config['queue_size'],
  File "/home/chad/codebase/project/src/sampler/NestedSampling.py", line 97, in sample
    sampler.run_nested(dlogz_init=dlogz, nlive_init=seed_num,
  File "/home/chad/codebase/project/venv/lib/python3.8/site-packages/dynesty-1.1-py3.8.egg/dynesty/dynamicsampler.py", line 1647, in run_nested
  File "/home/chad/codebase/project/venv/lib/python3.8/site-packages/dynesty-1.1-py3.8.egg/dynesty/dynamicsampler.py", line 1767, in add_batch
  File "/home/chad/codebase/project/venv/lib/python3.8/site-packages/dynesty-1.1-py3.8.egg/dynesty/dynamicsampler.py", line 1128, in sample_batch
  File "/home/chad/codebase/project/venv/lib/python3.8/site-packages/dynesty-1.1-py3.8.egg/dynesty/nestedsamplers.py", line 684, in update
[...]

PR #305, opened 2021-08-10
update error message if trying to get bounding ellipse of a single point

segasai:
Helps #302

### DIFF
--- a/py/dynesty/bounding.py
+++ b/py/dynesty/bounding.py
@@ -578,6 +578,9 @@ class MultiEllipsoid(object):
                              "to compute ellipsoid decompositions.")
 
         npoints, ndim = points.shape
+        if npoints == 1:
+            raise RuntimeError('Cannot compute the bounding ellipsoid of '
+                               'a single point.')
 
         # Calculate the bounding ellipsoid for the points, possibly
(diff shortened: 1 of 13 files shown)


segasai:
Hi, This branch is ready to be merged (it has a collection of various things that'd be nice to merge) 
I want to bring to particular attention this commit  54e268eeeb2d0e1551dbf2ff30d4020583d92199
as it fixes an annoying bug that I've introduced a couple of months ago in the update_interval calculation logic.

joshspeagle:
How is this looking? Can I looking into merging this in? I also want to push the fix to https://github.com/joshspeagle/dynesty/commit/54e268eeeb2d0e1551dbf2ff30d4020583d92199 that we missed and get the updated version on `pip` sooner rather than later.

segasai:
I think this is good to merge, with the only concern over the factor of two in front of the variance calculation (see another issue). If you are happy with that, I think this can be merged. (there is one more function where I'd like to reuse the compute_integrals(), but I can do it in a separate commit later)

joshspeagle:
Okay, then let me look into resolving that first, we can add it in, and then merge.

segasai:
Great! 
Thanks

Issue #307, opened 2021-08-18
Refactor iterators


segasai:
Currently various sampling iterators in the code return tuples 
`
yield (idx, ustar, vstar, loglstar, logvol, logwt, logz, logzvar,
                   h, 1, point_it, boundidx, bounditer, self.eff, delta_logz)
`
which are then unpacked
`
(worst, ustar, vstar, loglstar, logvol, logwt, logz, logzvar,
                 h, nc, worst_it, boundidx, bounditer, eff,
                 delta_logz) = results
 `
which is error-prone and pollutes the namespace with unused/confusing variables.

I think we need to do yield namedtuples and extract from them  just what we need. 
```
In [7]: NT=collections.namedtuple('R',['worst','ustar'])

In [8]: x=NT(worst=33,ustar=[3,5])

In [9]: x.worst
Out[9]: 33

In [10]: x.ustar
Out[10]: [3, 5]
```

joshspeagle:
100% agreed this would make things a lot cleaner.

Issue #308, opened 2021-08-18
KL divergence code is broken ? 


segasai:
I've tried to add a test of a KL divergence here 
https://github.com/segasai/dynesty/blob/95c4b09fe36a0c6bed3633b383b30c5581d3137d/tests/test_misc.py#L124
and it currently fails. 
```
        kl = dyutil.kl_divergence(res2, res1)
>       assert (kl[-1] > 0)
E       assert -0.056839523328596174 > 0
```
I disabled the sign check temporarily but it needs to be fixed (looked into).

joshspeagle:
Could this be the reason why?
https://github.com/joshspeagle/dynesty/blob/243025cbfc5f1941e9dc1b00348c54f8b877bd6c/py/dynesty/utils.py#L1202-L1207

segasai:
I don't quite know. I haven't investigated and the logic behind the code is not obvious to me.

joshspeagle:
In theory, it's just running through the calculation:
> kld = np.exp(logp1) * (logp1 - logp2)
> return np.cumsum(kld)

Since `res2` is just a bootstrap resampled version of `res1`, it occupies a subset of the points from `res1` but with different associated logXs (and therefore logwts). So the idea is to just go through and match up the points in `res1` with the closest corresponding point in `res2`, compute the KLD at the corresponding point, and then just sum them up across all the points. All the relevant lines look like they're implementing the right calculation
https://github.com/joshspeagle/dynesty/blob/243025cbfc5f1941e9dc1b00348c54f8b877bd6c/py/dynesty/utils.py#L1173
https://github.com/joshspeagle/dynesty/blob/243025cbfc5f1941e9dc1b00348c54f8b877bd6c/py/dynesty/utils.py#L1184
https://github.com/joshspeagle/dynesty/blob/243025cbfc5f1941e9dc1b00348c54f8b877bd6c/py/dynesty/utils.py#L1205-L1207
https://github.com/joshspeagle/dynesty/blob/243025cbfc5f1941e9dc1b00348c54f8b877bd6c/py/dynesty/utils.py#L1210

but there's probably something subtle happening that I'm just not seeing right now...

segasai:
I guess what doesn't quite make sense to me why the code need to look at sample_id's at all here. IMO the match need to be just by position. And all the points need to be binned by repeats with repeats logp summed and after that the two equal length logp arrays compared.
[...]

PR #309, opened 2021-08-20
Namedtuple iterator change

segasai:
Although it could have been an option after 1.2, I thought addressing #307 is worthwhile as currently things are error-prone. 
now all the iterators return namedtuples and I think the code is cleaner. There is a fair bit of repeated code stuff and it may be refactored in the future, but it looks worth applying. 

I've also yanked the kl_divergence.

### DIFF
--- a/py/dynesty/bounding.py
+++ b/py/dynesty/bounding.py
@@ -633,7 +633,7 @@ class MultiEllipsoid:
         # using Monte Carlo integration.
         if mc_integrate:
             self.logvol, self.funit = self.monte_carlo_logvol(
-                return_overlap=True)
+                rstate=rstate, return_overlap=True)
 
 
 class RadFriends:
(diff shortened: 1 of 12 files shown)


segasai:
This is ready to be reviewed after the tests pass (they take longer now) 
This PR contains
* The namedtuple fix
* Removal of kl_divergence
* adding rstate to some places in bounding where it was forgotten
* Moving bootstrap logic into a separate function that processes enlarge/boostrap information and makes the decision (right now it sets 1.25 enlarge by default, unless sampler='unif' in that case it sets bootstrap=5)
* While making the previous change I noticed that nestedsamplers.py has a bunch of code that is just copy-pasted 5 times. I've got rid of that, by creating another superclass from which all the classes inherit. That got rid of ~ 500 lines of code. 
* Some fixesto the pool testing code as it was running a bit too slow.

It fixes #307 , #308 and deals somewhat with #310 (TBC)

segasai:
Thanks

Issue #311, opened 2021-08-23
Weight function output


segasai:
While looking at this https://github.com/joshspeagle/dynesty/issues/240 issue I tried a simple code 

```
import test_dyn
import dynesty

sampler = dynesty.DynamicNestedSampler(test_dyn.loglike_egg,
                                       test_dyn.prior_transform_egg,
                                       2,
                                       nlive=500)
sampler.run_nested(maxcall=1000, use_stop=False)
sampler.add_batch(nlive=250, maxcall=1000)  # use_stop=False)

```

which gave the error 
Traceback (most recent call last):
  File "xx.py", line 9, in <module>
    sampler.add_batch(nlive=250, maxcall=1000)  # use_stop=False)
  File "/home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/dynamicsampler.py", line 1810, in add_batch
    for cur_results in self.sample_batch(nlive_new=nlive,
  File "/home/skoposov/pyenv38/lib/python3.8/site-packages/dynesty/dynamicsampler.py", line 1122, in sample_batch
    raise RuntimeError(
RuntimeError: ('Could not find live points in the required logl interval. Please report!\nDiagnostics. logl_min: 242.29591740055105 ', 'logl_bounds: (242.29591740055105, 242.29591740055105) ', 'saved_loglmax: 242.29591740055105')

Which is the same as 
this one https://github.com/joshspeagle/dynesty/issues/302

Which led me to think more about this. And I think what's happening here is when we get the results of the run that is 'incomplete' (i.e. stopped due to maxcall/maxiter or whatever) then it's possible that the weight  will be concentrated in the top live-point, so the logl_bound will be just the top point (+ pad or so).  And this is clearly an issue as it'll be hard to generate proposal points from that. 
So I'm wondering if we should  enforce that the interval covers at least nlive points ? (I.e. ideally I'd get rid of pad parameter and just use nlive to pad the interval if needed.

segasai:
Also my suggestion will automatically mean that if we just interrupted the run_nested before logz is reached, the add_batch would just resume with the top live points which is probably a good idea

joshspeagle:
Hm. This seems like a reasonable explanation and suggested fix, although I'd want to see how it modifies the overall behaviour and sampling efficiency in more typical use cases. It probably would be good to keep the padding in for consistency, so that the final size of the interval is essentially max(current padding scheme, nlive points) or something.

segasai:
[...]

PR #313, opened 2021-08-23
Batch fix

segasai:
Fix of #311 and #302

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -155,7 +155,7 @@ def weight_function(results, args=None, return_weights=False):
     logl = results.logl
     if bounds[1] > nsamps - 1:
         # overflow on the RHS, so we move the left side
-        bounds = [bounds[0] - (bounds[1] - nsamps - 1), nsamps - 1]
+        bounds = [bounds[0] + (bounds[1] - (nsamps - 1)), nsamps - 1]
     if bounds[0] < 0:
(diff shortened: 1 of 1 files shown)


segasai:
Okay, this is a rehash of https://github.com/joshspeagle/dynesty/issues/244 apparently. 
So my fix is incorrect, because we need to update logl_min, logl_max otherwise we'll fail to sample when some of our live points are below the threshold. 
Also, I realized that the weighting from here https://github.com/joshspeagle/dynesty/issues/244  which was coded to be  1/L_i 1/W_i  ( see also https://github.com/joshspeagle/dynesty/pull/248#issuecomment-845250127 ) is wrong..
If I understand correctly now weighting need to be X_i, i.e we need to downweight proportionally to volume occupied.

segasai:
I've tentatively addressed the issue. 
I've fixed the weights to be volumes. And if there are not enough points above the boundary I move the boundary. (updating the logl_min of the run). That rubs me the wrong way somewhat, because we are updating the inputs given to us, but maybe that's fine. Maybe we need to give the warning in this case. 
Also, apparently in the fix of the padding inside weight_function that I've done before (when if there is only one point on the right edge, I didn't expand it as correctly to the left ( that's why my very first hacky fix of #302 didn't work)

joshspeagle:
Okay, this looks good. I think the combination of the weight update and the more robust logic should hopefully now address the edge case where sampling terminates really early.

Issue #310, opened 2021-08-20
Failures in tests


segasai:
As a test of latest changes I've run the test suite in a loop with different seeds. 

and for example for these tests 
```
for a in `seq 0 35` ; do env  DYNESTY_TEST_RANDOMSEED=$a PYTHONPATH=py:tests:$PYTHONPATH pytest tests/test_dyn.py    > /tmp/log.${a} &  done 
```

there are ~ 15% of failures like this caused by the deviations being larger than 5 sigma 

```python
============================= test session starts ==============================
platform linux -- Python 3.8.10, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/skoposov/curwork/dynesty
plugins: cov-2.12.1, parallel-0.1.0
collected 1 item                                                               

tests/test_dyn.py F                                                      [100%]

=================================== FAILURES ===================================
___________________________________ test_dyn ___________________________________

    def test_dyn():
        # hard test of dynamic sampler with high dlogz_init and small number
        # of live points
        ndim = 2
        bound = 'multi'
        rstate = get_rstate()
        sampler = dynesty.DynamicNestedSampler(loglike_egg,
                                               prior_transform_egg,
                                               ndim,
                                               nlive=nlive,
                                               bound=bound,
                                               sample='unif',
                                               rstate=rstate)
        sampler.run_nested(dlogz_init=1, print_progress=printing)
        logz_truth = 235.856
>       assert (abs(logz_truth - sampler.results.logz[-1]) <
                5. * sampler.results.logzerr[-1])
E       assert 0.29720712094015767 < (5.0 * 0.04555387445144435)
E        +  where 0.29720712094015767 = abs((235.856 - 236.15320712094015))

tests/test_dyn.py:39: AssertionError
=========================== short test summary info ============================
FAILED tests/test_dyn.py::test_dyn - assert 0.29720712094015767 < (5.0 * 0.04...
========================= 1 failed in 75.96s (0:01:15) =========================
```

Presumably there are more.
And also It's probably worth rerunning this https://github.com/joshspeagle/dynesty/issues/289 test. 
I haven't yet decided what's the strategy here. I think my preferred strategy is to update the defaults so that we still fall within 5 sigma, rather than bumping up the thresholds. (in fact there are a few places in the tests already where the thresholds are pretty large).

joshspeagle:
Agreed on both counts.

segasai:
[...]

PR #315, opened 2021-08-23
Test fixes2

segasai:
Update a couple of easy tests to address #310 (test_exc and test_ellipsoids)
(branched out of master)


joshspeagle:
Okay, so should I expect to see some of the other possible fixes/issues discussed in #310 added in here as well?

segasai:
No. These were easy pickings. The rest is harder. It's better to merge those in

PR #316, opened 2021-08-24
Test fixes3

segasai:
This addresses most of the #310 
Now there are just 
 3 FAILED tests/test_dyn.py::test_dyn

in the run of 100 tests (x75 each) . 
I've also updated the boostrap logic
and added my test_code that allows diagnosing the quality of the ellipsoidal representation.


segasai:
thanks

Issue #186, opened 2020-05-11
merge_runs fails due to nlive not set and negative logzvar


ahallcosmo:
Hi,

I've encountered a couple of issues in the merge_runs function. Firstly it fails at the line

r.append(('logzerr', np.sqrt(np.array(combined_logzvar))))

since some values of my combined_logzvar are negative (presumably due to mis-estimation of the variance of logz).

Secondly, the merged results instance has not had its 'nlive' attribute set, so, e.g., results.summary() fails.

Cheers,
Alex

joshspeagle:
The first line

> np.sqrt(np.array(combined_logzvar))

should throw warnings but still go through. The `nan` values are indeed expected and due to the approximation scheme used to estimate the variance. A more robust method for estimating the variance can be achieved through some of the functions in `dynesty.utils` as outlined in the docs.

The second issue

> the merged results instance has not had its 'nlive' attribute set, so, e.g., results.summary() fails.

is indeed an actual bug that's been raised in #175 and #184. It should be a small fix but I just haven't gotten around to it yet.

Issue #184, opened 2020-05-09
DynamicNestedSampler.results.summary function fails because of nlive


JohannesBuchner:
I am not sure why, but the following code fails for me at the last line:

```#!/usr/bin/env python
# -*- coding: utf-8 -*-
from __future__ import (print_function, division)
import sys
import numpy as np
import dynesty

# EGGBOX
def loglike(x):
    tmax = 5.0 * np.pi
    t = 2.0 * tmax * x - tmax
    return (2.0 + np.cos(t[0] / 2.0) * np.cos(t[1] / 2.0)) ** 5.0


def prior_transform(x):
    return x


print("with nested sampler:")
sampler = dynesty.NestedSampler(loglike, prior_transform, 2)
sampler.run_nested()
sampler.results.summary()

print("with dynamic nested sampler:")
sampler = dynesty.DynamicNestedSampler(loglike, prior_transform, 2)
sampler.run_nested()
sampler.results.summary()
```

with the output:

```
± python3 run1.py 
with nested sampler:
3851it [00:06, 560.91it/s, +500 | bound: 7 | nc: 1 | ncall: 22127 | eff(%): 19.664 | loglstar:   -inf < 243.000 <    inf | logz: 236.150 +/-  0.153 | dlogz:  0.001 >  0.509]
Summary
=======
nlive: 500
niter: 3851
ncall: 22127
eff(%): 19.664
logz: 236.150 +/-  0.153
with dynamic nested sampler:
21297it [05:42, 62.15it/s, batch: 11 | bound: 57 | nc: 3 | ncall: 55311 | eff(%): 38.504 | loglstar: 235.513 < 243.000 < 242.767 | logz: 235.858 +/-  0.157 | stop:  0.920]        
Traceback (most recent call last):
  File "/mnt/data/daten/PostDoc2/home/dynesty/dynesty/results.py", line 181, in __getattr__
    return self[name]
KeyError: 'nlive'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "run1.py", line 34, in <module>
    sampler.results.summary()
  File "/mnt/data/daten/PostDoc2/home/dynesty/dynesty/results.py", line 205, in summary
    .format(self.nlive, self.niter, sum(self.ncall),
  File "/mnt/data/daten/PostDoc2/home/dynesty/dynesty/results.py", line 183, in __getattr__
    raise AttributeError(name)
AttributeError: nlive
```

This is with dynesty 1.0.2.

[...]

Issue #175, opened 2020-02-26
Model averaging


MrBudgens:
Am I right in thinking that dynesty.utils.merge_runs can be used to perform Bayesian model averaging? Specifically, if I use two or more different models to infer the same parameters from the same data using the same priors, can their dynamic sampling runs be trivially merged to marginalise over the models? If so, may I suggest a word or two be added to the documentation to make this explicit, with advice on how to assign non-equal weights to models. If not, what have I missed?

Also, when I try to use merge_runs I can't use .summary to see the results (using .quantile still works):

Traceback (most recent call last):
  File "...dynesty\results.py", line 181, in __getattr__
    return self[name]
KeyError: 'nlive'

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "<ipython-input-48-68797f792eca>", line 1, in <module>
    merged.summary()
  File "...\dynesty\results.py", line 205, in summary
    .format(self.nlive, self.niter, sum(self.ncall),
  File "...\dynesty\results.py", line 183, in __getattr__
    raise AttributeError(name)

AttributeError: nlive

joshspeagle:
> Am I right in thinking that dynesty.utils.merge_runs can be used to perform Bayesian model averaging?

I had not considered this possibility. My first thought is I do not believe this will work: the procedure used to merge the runs together assumes that they trace the same underlying distribution (i.e. that tracking the relative log-likelihoods between the corresponding runs can be used to determine their combined positions with respect to the prior volume), so if this assumption is violated I don't know if the same results would hold. Have you run any simple tests with, e.g., mixing various Gaussian likelihoods that would at least make this seem plausible? I don't know if I've seen this result published anywhere, so if this is true in at least some ideal cases it'd be worth following up more rigorously.

> Also, when I try to use merge_runs I can't use .summary to see the results (using .quantile still works):

Ahhh, this is because after `merge_runs` is applied I change all runs to be considered "dynamic" internally. I don't think I ever quite got around to getting the `results.summary()` call to function appropriately in that case, but it's a small enough feature request I should just add in that functionality for consistency anyways. I've flagged it and added it to my to-do list. Thanks!

MrBudgens:
Based on exactly one attempt, the posteriors I merged produced a believable-looking result, but testing with Gaussians seems a better starting point for a real test - I'll investigate. I'm very much a beginner here, but as an alternative it seems to me that samples generated from each model run using resample_equal could be combined to achieve my aim?

[...]

PR #330, opened 2021-09-10
New results object

segasai:
Here's a working preliminary implementation of the Results() that is more than just a dictionary. 
It still accepts the list of key values or a dictionary, but prevents manual updates, and adds some additional methods

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -552,8 +552,6 @@ class DynamicSampler:
 
         state = self.__dict__.copy()
 
-        del state['rstate']  # remove random module
-
         # deal with pool
         if state['pool'] is not None:
             del state['pool']  # remove pool
@@ -562,7 +560,6 @@ class DynamicSampler:
         # deal with internal sampler (to be safe)
         if state['sampler'] is not None:
             try:  # attempt to remove the same things
-                del state['sampler'].rstate
                 if state['sampler'].pool is not None:
                     del state['sampler'].pool
                     del state['sampler'].M
--- a/py/dynesty/results.py
+++ b/py/dynesty/results.py
@@ -6,6 +6,7 @@ Utilities for handling results.
 """
 
 import sys
+import copy
 import numpy as np
 import shutil
 
@@ -227,13 +228,25 @@ def print_fn_fallback(results,
 
 class Results:
     """Contains the full output of a run along with a set of helper
-    functions for summarizing the output."""
+    functions for summarizing the output.
+    """
     def __init__(self, key_values):
+        """
+        Initialize the results using the list of key value pairs
+        or a dictionary 
+        Results([('logl',[1,2,3]),('samples_it',[1,2,3])])
+        Results(dict(logl=[1,2,3],samples_it=[1,2,3]))
+        """
         self._keys = []
         self._initialized = False
-        for k, v in key_values:
+        if isinstance(key_values, dict):
+            key_values_list = key_values.items()
+        else:
+            key_values_list = key_values
+        for k, v in key_values_list:
+            assert (k not in self._keys)  # ensure no duplicates
             self._keys.append(k)
-            setattr(self, k, v)
+            setattr(self, k, copy.copy(v))
         required_keys = ['samples_u', 'samples_id', 'logl', 'samples']
         for k in required_keys:
             if k not in self._keys:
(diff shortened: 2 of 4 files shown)


joshspeagle:
I see there are some unresolved comments related to the results structure that is being saved. I've added some additional info below.

- `samples_it`: the sampling iteration when the sample was proposed (e.g., iteration 570)
- `samples_id`: the unique ID of the sample (e.g., ID=2)

These are primarily useful for "winding back" sampling if we want to see what the live points are doing at some intermediate time during the run, such as in `boundplot`.

- `samples_n`: the number of live points at the point when the sample was proposed (e.g. K=1957)

This is mainly useful for dynamic nested sampling.

- `samples`: the location (in original coordinates). Should be identical to `samples_v` I think.
- `eff`: the overall sampling efficiency (`niter`/`ncall`, more or less)

These are just summary aliases, IIRC.

- `bound`: the set of bounding objects used to condition proposals
- `bound_iter`: the iteration when the corresponding bound was created to propose new live points (e.g., iteration 520)
- `samples_bound`: the index of the bound that the corresponding sample was drawn from (e.g., bound 1)

These are useful when using plotting functions like `cornerbound`, since they allow us to track particular bounds. I also left them in in case I wanted to tackle adding something like Importance Nested Sampling into the code, which would require knowing which sample came from which bound.

- `samples_batch`: tracks the batch during which the samples were proposed
- `batch_bounds`: the log-likelihood bounds used to sample points in a given batch
- `batch_nlive`: the number of live points added in a given batch

These are mostly additional quantities I added to make sure I could associate points with batches and track that the batches were being added correctly.

Let me know if I missed anything.

segasai:
I've fixed some XXs, but some questions still remain remain regarding batch quantities and lengths (i.e. whether something has a lenght niter or not). It'd be good for you to edit this directly. 
Also I don't quite understand what's the difference between batch_nlive vs samples_n and samples_bound vs batch_bound. 
It'd be good to clarify that. Maybe even rename some of them for consistency. 

Also _merge_two()  function in the utils.py mentions prop_iter, samples_prop and prop properties, but those do not seem to be used anywhere. I assume those are dead ? 

The reason why I think this needs cleaning up, is because resutl is an externally returned object, as opposed to RunRecord, so it would be good to ensure, it is fully understood/described what each of those are. Plus one can establish some kind of tests that verify that arrays have correct lengths etc. ATM I'm not sure of that. 
Also I was potentially thinking of storing the information about proposed point for rwalk/rslice, so one could compute the degree of correlatedness between samples, but I wouldn't want to do that before the results object is fully described.

joshspeagle:
> I've fixed some XXs, but some questions still remain remain regarding batch quantities and lengths (i.e. whether something has a lenght niter or not). It'd be good for you to edit this directly.

That's easy enough for me to do post-merge. No problem.

> Also I don't quite understand what's the difference between batch_nlive vs samples_n and samples_bound vs batch_bound.
It'd be good to clarify that. Maybe even rename some of them for consistency.

`samples_n` tracks the number of effective live points at every iteration, e.g., [180, 179, 178, 177, 190, 190, 190, 195, ...]. `batch_nlive` just checks how many live points we added in from a particular batch, e.g., 250. `samples_bound` is the associated bounding distribution, whereas `batch_bound` is the `logl_bound` tuple. I agree that renaming both quantities (esp. the second) would be helpful, especially since I don't think they are currently used elsewhere in the code.

> Also _merge_two() function in the utils.py mentions prop_iter, samples_prop and prop properties, but those do not seem to be used anywhere. I assume those are dead ?

These were renamed to `bound`. I can try to update this as well when I go through the internal docstrings.

> Also I was potentially thinking of storing the information about proposed point for rwalk/rslice, so one could compute the degree of correlatedness between samples, but I wouldn't want to do that before the results object is fully described.

This would be great and get to addressing (in part) #163, since then we would have the possible option of using that info for tuning the number of walks/slices on the fly.

joshspeagle:
There's a small conflict with the main branch I need to double-check on, but I will try to merge this in tonight and then get around to cleaning everything up in preparation for the v1.2 release.

segasai:
Thanks.
The conflict can be resolved completely in favor of the results_branch.

segasai:
I think this probably closes the #175 and #186 too

joshspeagle:
Agreed. I tentatively say this:
- Resolves #175. 
- Resolves #184. 
- Resolves #186.

Issue #331, opened 2021-09-10
Usage of KLD for stopping criteria


segasai:
I've been looking at the stopping criteria for the dynamic sampling and I kind'a struggle to understand the rationale why the mean(KLD)/stddev(KLD) for KLD that's always positive is a meaningful statistic to optimize when deciding when to stop. 

If I understand correctly the most likely thing people will care about for the posterior is the effective number of samples, as I think that'll dominate the error-budget, as opposed to the volume misestimates (which is what the default stopping function with jitter_run tries to estimate)

Also in fact the 'weight_function' tries to essentially find the best logl range to maximize Neff (minimizing the maximum weights) , so therefor it's kind'a weird that the stop function is measuring something else.

joshspeagle:
I'm perfectly happy switching over the default choice of stopping criteria to some combination of `Neff` and `logzerr` instead of the current `KLD_SNR` and `logzerr` weighting. I believe that functionality was added to the code a while back, so it's an easy change to make.

PR #332, opened 2021-09-11
Stopping change

segasai:
get rid of KL divergence based stop function
Use the n effective sample as a metric instead. The default configuration requires 10000  samples.
see #331 
This is based on #329 branch to avoid merging conflicts

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -219,11 +219,10 @@ def stopping_function(results,
 
         stop_evid = evid_std / evid_thresh
 
-    The posterior stopping value is based on the fractional error (i.e.
-    standard deviation / mean) in the Kullback-Leibler (KL) divergence
-    relative to a given threshold::
+    The posterior stopping value is based on the estimated effective number
+    of samples.
 
-        stop_post = (kld_std / kld_mean) / post_thresh
+        stop_post = neff / target_neff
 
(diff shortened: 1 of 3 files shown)


joshspeagle:
This looks good, but before merging I need to request that the **old stopping function be included for backwards compatibility**, even if it is disabled/unused by default. One easy option is just to define it internally (and then I can provide an example in the docs showing users how to swap it back in to attain previous behaviour). Alternately, you could add a text-based option where passing in something like `'kld_error'` gives the old behaviour but `None` or `'n_eff'` gives the current one.

Does that sound reasonable to you?

segasai:
I think I agree, it's probably good to keep the stopping function for a release or two with the deprecation warning, but given that all the code is versioned and available, I'm reluctant to keep old 'dead' code around for long. (and if it's not tested, it'll be broken quickly). 
(i.e. we'd say that to restore previous stopping behaviour the user would need to 
do stopping_function=dynesty.utils.old_stopping_function, neff=None)

Also regarding the new stopping function, it's not quite clear to me iif t makes much sense to do MCMC resamples there (by default), as the logz error-bar is probably reasonably robust.

joshspeagle:
> Also regarding the new stopping function, it's not quite clear to me iif t makes much sense to do MCMC resamples there (by default), as the logz error-bar is probably reasonably robust.

That's fair. We can probably change the default option to just take the logzerr approximation, rather than recompute it from resampling. That should be a minor change.

Issue #230, opened 2021-03-30
uniform distribution can't be sampled


segasai:
Hi,

while adding tests I noticed that the uniform distribution fails to sample with current code. 
It looks like it's connected to this https://github.com/joshspeagle/dynesty/pull/203

```
import numpy
import dynesty

# seed the random number generator
np.random.seed(5647)

nlive = 1000
printing = True


def loglike(x):
    return 0

def prior_transform(x):
    return x


def test_sample():
    ndim = 2
    sampler = dynesty.NestedSampler(loglike,
                                    prior_transform,
                                    ndim,
                                    nlive=nlive,
                                    bound='multi',
                                    sample='unif')
    sampler.run_nested(dlogz=0.1, print_progress=printing)
    logz_truth = 0
    # this the integral
    logz, logzerr = sampler.results.logz[-1], sampler.results.logzerr[-1]
    thresh = 5
    assert ((logz - logz_truth) < thresh * logzerr)
```

I don't actually know what's the fix here.

segasai:
It looks like 
https://github.com/joshspeagle/dynesty/blob/858bb15dedcff600617cecb77fd6f3f2cc9fb3c9/dynesty/sampler.py#L379
here there is  basically an infinite loop when all the likelihood values are the same. 

Also, I remember reading something in arxiv about this: https://arxiv.org/abs/2010.13884
So apparently in multinest there are  tiny perturbations to logl values that resolve that issue (which are technically not quite right)

joshspeagle:
Yes, this is a fundamental issue with nested sampling -- it can't deal with likelihood slabs (it requires a unique mapping from a particular likelihood value L to a particular prior volume X, which is broken if this is a *region* rather than a *contour*). The tiny perturbations *might* be a practical fix to this problem, but it's an edge case for sure.

segasai:
I think one potential way of dealing in practice is to at least raise a warning if all the likelihood values of live-points are exactly the same and proposal fails for many iterations to alarm the user of the problem.

joshspeagle:
Yea, this is a good idea. I'll put a note down to add a check, flag, and warning for users if it looks like we've hit this particular edge-case.

a-lhn:
I'm having a problem where I think it's possible that there are regions of similar likelihood values causing difficulties with sampling. I have looked through the code but I'm not sure I understand how to return and check the likelihood values of live points to diagnose the problem; would it be possible for any advice on how to go about this until the fix is implemented? Thanks

[...]

Issue #242, opened 2021-04-26
Add handling of likelihood plateaus?


mattpitkin:
Is anyone planning on implementing the handling of likelihood plateaus suggested in https://arxiv.org/abs/2010.13884? It looks like implementing their Algorithm 2 would be fairly simple and backwards compatible. I can give it a go if it's something that is wanted.

joshspeagle:
Sorry for the delay in getting back to this. As far as I know, no one is currently planning on implementing this procedure. I think it's straightforward enough in theory (if the acceptance criterion is swapped back to `>=` instead of `>`), so if you'd like to tackle adding this in I'd definitely appreciate it!

mattpitkin:
No problem. Hopefully I'll find some time to give this a go.

PR #333, opened 2021-09-14
put a warning if all likelihood values are the same. 

segasai:
Partially address #230


joshspeagle:
Will consider this commit to temporarily resolve #230 and resolve #242. I can re-open those if we would prefer to try and add in specific checks for dealing with plateaus (adding those live points in a batch update rather than just one at a time, more or less), but just flagging things with warnings should be okay for now.

Issue #320, opened 2021-08-30
rwalk/rstagger issues 


segasai:
While running a few more checks for rwalk/rstagger, 
Specifically the dimensionality test:

python -c 'import test_highdim as T; T.do_gaussians(sample="rwalk")'

Here is a bias vs dimension for different samplers (all other parameters are default)

(updated image)
![image](https://user-images.githubusercontent.com/109513/131360869-cc55c1e6-f0af-440d-84b3-018a375e32a1.png)

I'm seeing that for rslice/unif samplers things are good, but for the rwalk the bias creeps in vs dimension. 
(to see if this is caused by my recent changes, I've also checked this 9b8d3f1620fbc06b178aa30cd529967540447902  (shown in red). And I think it shows we need to increase the walks parameter. 

Also rstagger seems broken and I can't see the same problem in 9b8d3f1620fbc06b178aa30cd529967540447902 so it's something broken by a most recent change.

segasai:
A somewhat connected question is 

What is the rationale behind this logic in update_rwalk
https://github.com/joshspeagle/dynesty/blob/b030ff76246b9b9a62c3e4da1285d8667d553fe5/py/dynesty/nestedsamplers.py#L152

It seems pretty involved without much justification.

segasai:
The #321 restores the previous behaviour of rstagger. But it is still pretty bad.

![image](https://user-images.githubusercontent.com/109513/131377120-64ffffaf-f576-41d2-a3ff-b9f6c4eac424.png)

I think there is something not right from statistical point of view in rstagger. I thing it's incorrect in general as it doesn't  have the right invariant distribution.
In fact here is a test program that is supposed to be sampling uniformly withing a diamond using stagger approach:

```
import numpy as np
import dynesty.sampling as S

def logl(x0):
    x, y = x0
    a = 10
    sl = 0.01
    x1 = np.abs(x)
    if x1 > a:
        return -1

    if np.abs(y) < (sl * (x1 - a)**2):
        return 1
    else:
        return -1


def priort(x):
    return x * 20 - 10

logls = 0
axes = np.eye(2) * .1
scale = 1
rstate = np.random.default_rng()
kwargs = {}
u = [.5, .5]
print(logl(priort(np.r_[0.5, 0.5])))
ret = []
for i in range(10000):
    u = S.generic_random_walk(u,
                              logls,
                              axes,
                              scale,
                              priort,
                              logl,
                              rstate,
                              kwargs,
[...]

PR #354, opened 2022-02-11
return random point if no MCMC steps accepted

ColmTalbot:
Hi @joshspeagle @segasai this PR fixes an issue in the `rwalk` proposal method.

Currently, if no steps are accepted the original point is returned. However, this point will almost always be accepted as there's only a 1 / nlive probability that it doesn't have a higher likelihood than the lowest likelihood live point (by definition).

Having repeated nested samples in the chain can easily lead to subtle biases and overly constrained ellipses.

The proposed fix is to just return a random point from the prior. This will almost certainly not be accepted but communicates to the sampler that no valid point was found.

(FWIW, we've been using a [version of this](https://git.ligo.org/lscsoft/bilby/-/blob/master/bilby/core/sampler/dynesty.py#L867-L870) in `Bilby` for a while.)

### DIFF
--- a/py/dynesty/sampling.py
+++ b/py/dynesty/sampling.py
@@ -257,9 +257,9 @@ def generic_random_walk(u, loglstar, axes, scale, prior_transform,
         else:
             nreject += 1
     if naccept == 0:
-        # Technically we can find out the likelihood value
-        # stored somewhere
-        # But I'm currently recomputing it
+        # No MCMC steps were accepted. To avoid repeated nested
+        # samples return a random draw from the prior.
+        u = rstate.uniform(0, 1, n)
         v = prior_transform(u)
         logl = loglikelihood(v)
 


segasai:
Hi @ColmTalbot ,

Thanks for the patch,  but I am not sure that's statistically correct.
MH random walk requires a very specific proposal distribution to keep the distribution invariant and preserve the detailed balance. The current  walker certainly does that (at the cost of possibly repeated points). I am not convinced that your code does that. 

Also I agree the behaviour of the rwalk sampler when none of the samples are accepted is potentially problematic, but  I am not convinced that your proposed solution is a correct one.

segasai:
The reason why I'm concerned about this code is because the previous implementation (in 1.1)  was exactly incorrect here. And it was showing up in cases where your posterior has very sharp corner on one side vs another, so that the acceptance rate on one side is much lower. You can't just give on on a proposed live point and try another one -- that'll lead to incorrect posterior.

segasai:
I agree that we need to communicate that the point evolution was not successful, but the way you propose I believe will not be correct, because the most likely outcome of that will be that the point will be rejected and some other live-point will likely be selected as a new candidate, but that's not correct.


See the example from https://github.com/joshspeagle/dynesty/issues/320
I.e. imagine cases where the acceptance probability from rwalk is small on one side and very large on another. If we fail in rwalk, we at least MUST ensure that we start from the same life-point (maybe after adjusting the parameters of the proposal), but even then is a bit worrisome. 

In fact I took your code and ran it using the example from #320

```
import numpy as np
import dynesty.sampling as S


def logl(x0):
    x, y = x0
    a = 10
    sl = 0.01
    x1 = np.abs(x)
    if x1 > a:
        return -1

    if np.abs(y) < (sl * (x1 - a)**2):
        return 1
    else:
        return -1


def priort(x):
    return x * 20 - 10


logls = 0
axes = np.eye(2) * .1
scale = 1
rstate = np.random.default_rng()
kwargs = {}
u = [.5, .5]
print(logl(priort(np.r_[0.5, 0.5])))
ret = []
for i in range(100000):
    u = S.generic_random_walk(
        u,
        logls,
        axes,
        scale,
        priort,
        logl,
        rstate,
        kwargs,
    )[0]
    ret.append(u)
ret = np.array(ret)

```
And I'm not getting a uniform distribution inside the diamond like I should 

image

(the existing dynesty code (from master) does behave correctly)


So 
I think first and foremost I think we should keep generic_random_walk intact because it implements a classical MH step and should do one simple thing. 
I agree that we may need to consider some logic in _fill_queue() probably for the case when we get repeats, but we very carefully need to think and test it.

(Also I or you can be wrong in these arguments, that why I'd prefer to see a few actual test-cases here, like the one above, so we can actually verify what we are talking about and the code is correct, as this is critical component of the code).

segasai:
Maybe an even more concrete example. 

I'm worried about cases where our live points are distributed like this

```
xs,ys=2*np.random.uniform(size=(2,200))-1;
ys[xs>0]=ys[xs>0]/30.;
xs[xs>0]=30*xs[xs>0]
```
image

with say unit Gaussian proposal. 
If we allow to not accept gaussian walks from the narrow part of the distribution (because they'll rarely work), than the narrow part of the distribution will just die off, which is not correct ( as it has the same volume).

Obviously this is somewhat a corner case, but I don't think it's that extreme.

segasai:
I am not sure we are converging here. 

I think increasing number of steps just hides the problem by making more likely that the steps will be accepted; If they are there is no problem , the problem occurs if they are *NOT* accepted, therefore we should be testi

And suggestions you give for a hammer problem are good ones but they are not solving the underlying issue, that the sampling is incorrect. 

It could easily be the case that I'm in the wrong here, but because we discussing in the abstract, it's going to be hard to convince what's better. 

For example this is the code that samples the 'hammer' posterior using full dynesty dynamic sampler

```
import dynesty
import numpy as np
import dynesty.utils as dyfunc


def logl(X):
    x, y = X
    r = 1e-3
    scale = 30
    ret1 = -0.5 * (x**2 + y**2) / r**2
    ret2 = -0.5 * ((x / scale)**2 + (y / (1. / scale))**2) / r**2
    return ret1 * (x < 0) + ret2 * (x >= 0)


def cube(x):
    return 2 * x - 1


def doit(sample, bound='single'):
    rstate = np.random.default_rng(1)
    samp = dynesty.DynamicNestedSampler(
        logl,
        cube,
        2,
        nlive=100,
        bound=bound,
        sample=sample,
        # 'rwalk',
        # sample='rslice',
        rstate=rstate)
    samp.run_nested(n_effective=20000)
    samples = samp.results.samples
    wt = np.exp(samp.results.logwt - samp.results.logz[-1])
    samples_equal = dyfunc.resample_equal(samples, wt)
    return samples_equal
````

And here I compare the rslice, multi results (assume that's the truth; black) vs rwalk, single (red histogram). 

Here is the dynesty/master 
image

And here's with your patch
image

I'd argue that even default dynesty/master is pretty bad, but with the patch the agreement is even worse. 

So I'm not convinced that this an improvement. 

* The reason why it's still useful to test with 'single' bounding is because in high dimensions it's very easy to reach the point where you can't split your ellipsoid more due to numerical constraints (takes ~ 60 points to define 10dim ellipsoid) , so ensuring good 'single' behaviour is important.

**Comment by segasai** [comment]
It's great that we did converge. From my point of view the rslice sampler is much more stable because it's guaranteed to produce a sample always (at the cost of needing more steps for full mixing).
In fact one possibility is to utilise rslice as a fall-back option for rwalk (although one has to think about detailed balance)

**Comment by segasai** [comment]
For what it is worth, here's a patch implementing  the idea that in the case of failed rwalk step we do an rslice step instead 
```patch
diff --git a/py/dynesty/sampling.py b/py/dynesty/sampling.py
index 9e60d2a..ce0f15b 100644
--- a/py/dynesty/sampling.py
+++ b/py/dynesty/sampling.py
@@ -148,8 +148,14 @@ def sample_rwalk(args):
     (u, loglstar, axes, scale, prior_transform, loglikelihood, rseed,
      kwargs) = args
     rstate = get_random_generator(rseed)
-    return generic_random_walk(u, loglstar, axes, scale, prior_transform,
-                               loglikelihood, rstate, kwargs)
+    ret = generic_random_walk(u, loglstar, axes, scale, prior_transform,
+                              loglikelihood, rstate, kwargs)
+    if ret[-1]['accept'] == 0:
+        warnings.warn('Random walk failed to generate a single accepted point.'
+                      ' Generating a proposal using rslice sampler instead')
+        ret1 = sample_rslice(args)
+        return ret1[0], ret1[1], ret1[2], ret1[3] + ret[3], ret[4]
+    return ret
```

This will not have repeated samples, but also for my test problem doesn't really improve the posterior, so I don't 
really want to apply this patch as it is. I think what's needed is some tracking of correlation between samples.

PR #357, opened 2022-03-01
Use logz directly when `logplot=True` to avoid overflows

vandalt:
Hello,

I was facing a case where the final `logz` is high (around 2000). Everything works fine, but when I try to generate the `runplot()`, I get a NaN/Inf matplotlib error (see full error below). I tried using the `logplot=True` option, but because it does not use the log directly and still uses `np.exp(logz)` before scaling the axes in log, the result was the same.

I tried to use logz directly in the plot when `logplot=True` and it did fix the issue. The changes are similar to what is done in the [ultranest version of this function](https://github.com/JohannesBuchner/UltraNest/blob/master/ultranest/plot.py#L139).

Thank you!

----

Full error message (which is fixed by the PR when settings `logplot=True`):
```
/home/vandal/astro/dynesty/py/dynesty/plotting.py:201: RuntimeWarning: overflow encountered in exp
  data = [nlive, np.exp(logl), np.exp(logwt), logz if logplot else np.exp(logz)]
/home/vandal/astro/dynesty/py/dynesty/plotting.py:225: RuntimeWarning: overflow encountered in exp
  zspan = (0., 1.05 * np.exp(logz[-1] + 3. * logzerr[-1]))

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [4], in <module>
      1 from dynesty import plotting as dyplot
      2 reload(dyplot)
----> 4 rfig, raxes = dyplot.runplot(results, logplot=False)

File ~/astro/dynesty/py/dynesty/plotting.py:261, in runplot(results, span, logplot, kde, nkde, color, plot_kwargs, label_kwargs, lnz_error, lnz_truth, truth_color, truth_kwargs, max_x_ticks, max_y_ticks, use_math_text, mark_final_live, fig)
    259         ymax = max(span[i][1], yspan[i][1])
    260     axes[i].set_xlim([xmin, xmax])
--> 261     axes[i].set_ylim([ymin, ymax])
    263 # Plotting.
    264 labels = [
    265     'Live Points', 'Likelihood\n(normalized)', 'Importance\nWeight',
    266     'log(Evidence)' if logplot else 'Evidence'
    267 ]

File ~/astro/phy3051-6051/venv/lib64/python3.10/site-packages/matplotlib/axes/_base.py:4019, in _AxesBase.set_ylim(self, bottom, top, emit, auto, ymin, ymax)
   4017 self._process_unit_info([("y", (bottom, top))], convert=False)
   4018 bottom = self._validate_converted_limits(bottom, self.convert_yunits)
-> 4019 top = self._validate_converted_limits(top, self.convert_yunits)
   4021 if bottom is None or top is None:
   4022     # Axes init calls set_ylim(0, 1) before get_ylim() can be called,
   4023     # so only grab the limits if we really need them.
   4024     old_bottom, old_top = self.get_ylim()

File ~/astro/phy3051-6051/venv/lib64/python3.10/site-packages/matplotlib/axes/_base.py:3605, in _AxesBase._validate_converted_limits(self, limit, convert)
   3602 converted_limit = convert(limit)
   3603 if (isinstance(converted_limit, Real)
   3604         and not np.isfinite(converted_limit)):
-> 3605     raise ValueError("Axis limits cannot be NaN or Inf")
   3606 return converted_limit

ValueError: Axis limits cannot be NaN or Inf
```

### DIFF
--- a/py/dynesty/plotting.py
+++ b/py/dynesty/plotting.py
@@ -195,7 +195,7 @@ def runplot(results,
             mark_final_live = False
 
     # Determine plotting bounds for each subplot.
-    data = [nlive, np.exp(logl), np.exp(logwt), np.exp(logz)]
+    data = [nlive, np.exp(logl), np.exp(logwt), logz if logplot else np.exp(logz)]
     if kde:
         # Derive kernel density estimate.
         wt_kde = gaussian_kde(resample_equal(-logvol, data[2]))  # KDE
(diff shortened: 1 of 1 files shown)


segasai:
Sorry for the delay. 
I merged it (after adding a test with logplot=True)
Thank you for the contribution

PR #365, opened 2022-04-07
Allow variable mcmc length

ColmTalbot:
This PR adds a new option that allows the length of the MCMC exploration for the `rwalk` method to vary during the analysis such that a target number of proposed steps are accepted. There is also an option to disable the scale adaptation of the proposal.

- The default behaviour matches the existing behaviour.
- There is a new unit test with the variable MCMC length that passes well.

cc @joshspeagle @segasai

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -606,7 +606,8 @@ class DynamicSampler:
         d = {}
         for k in [
                 'nc', 'v', 'id', 'batch', 'it', 'u', 'n', 'logwt', 'logl',
-                'logvol', 'logz', 'logzvar', 'h', 'batch_nlive', 'batch_bounds'
+                'logvol', 'logz', 'logzvar', 'h', 'batch_nlive', 'batch_bounds',
+                'scale', 'walks',
         ]:
             d[k] = np.array(self.saved_run.D[k])
 
@@ -633,6 +634,7 @@ class DynamicSampler:
             results.append(
                 ('samples_bound', np.array(self.saved_run.D['boundidx'])))
             results.append(('scale', np.array(self.saved_run.D['scale'])))
+            results.append(('walks', np.array(self.saved_run.D['walks'])))
 
         return Results(results)
 
@@ -865,7 +867,9 @@ class DynamicSampler:
                                 n=self.nlive_init,
                                 boundidx=results.boundidx,
                                 bounditer=results.bounditer,
-                                scale=self.sampler.scale)
+                                scale=self.sampler.scale,
+                                walks=self.sampler.walks,
+                                )
 
                 self.base_run.append(add_info)
                 self.saved_run.append(add_info)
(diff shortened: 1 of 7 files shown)


segasai:
Thanks for the PR @ColmTalbot 

I think for this to be merged I'd like to see.
* an example where this new behaviour performs better than the existing behaviour with similar parameters.

I can imagine the new behaviour is better
* always (then we want to replace what we have by this new scaling)
* never, then we don't want to do anything
* for some distribitutions. Then it'd be good to have some idea what are those. 

but at least we need one example so we can assess.

segasai:
> To clarify, the first example above is a 15-dimensional bimodal Gaussian (I forgot to say that above.)

I see, that makes more sense. Thanks. Was it with a single ellipsoidal bound ?

segasai:
> To address the first point. If we don't match the correct distribution, we don't have an independent sample. That's the definition of the "correct" distribution. I think the question we normally ask is "how correlated can the new sample be with the old sample and not be visibly biased", which can be answered by running a bunch of simulations for the problem at hand, but I don't know of a general rule.

Just a quick answer on that. 
What I meant by 'correct' vs independent is that if the sampling scheme is violating detailed balance and/or doesnt' have the right invariant distribution I call it incorrect and the whole thing is broken. The indepedence of samples is indeed something that is never achieved exactly and will lead to biases, but the hope that they are indepedent "enough" or the most independent we can afford for a given CPU-time budget and the biases are small enough.

segasai:
Regarding the ACF, I haven't used it much, so I am not too familiar with its limitations, so you may be right there. My thinking was that we don't need a very accurate measure of it, we just need to detect cases where the samples are too correlated and we need to 'walk' more. So I thought that there should be enough information on that between each update_rwalk() call. 
But again, maybe it is unfeasible. 

Regarding the distance distribution, I can see your point now better, that difference between distribtuion of distances and the "true" distribution of distances may be an easy to compute metric to quantify the independence of samples. 

Given that at each point of the sampling we have N-points that are presumably samples from L>L_* 
We can assume that the pairwise distances should be a fair sample of the "true" distribution, so maybe we should use that as an indication that we need to have more walks. I.e. if the median separation of the live-points is x, we may want to increase walks till the typical separation is some fraction of x ?

I've found this case of the  high-dim shell quite instructive -- 

```python
import dynesty
import numpy as np
import scipy.stats

nlive = 1000
s0 = 0.01

ndim = 10


def loglike(X):
    return -100 * np.log(1 + np.abs((np.sqrt((X**2).sum()) - 1) / s0)**2)


def prior_transform(x):
    return x * 10 - 5

sampler = dynesty.NestedSampler(loglike,
                                    prior_transform,
                                    ndim,
                                    nlive=nlive,
                                    bound='single',
                                    rstate=rstate,
                                    sample='rwalk',
                                    walks=1)
    sampler.run_nested(print_progress=True)

```

with this behaviour of the scale parameter 
image

which clearly leads to a very small step size. 
here's the x_initial vs x_after_the_walk 
image

and here is the walking distance as a function of iteration. 
image
while for a 10d sphere with a radius of 0.1 the distance between two random points should be ~ 0.15. 

So I think this is the perfect test case to remedy. We would want I think, keep the same 'scale' as we have now, but we would want the walks to increase in the end of the iterations. 
I haven't tried with your patch what actually happens.

PR #384, opened 2022-08-16
improve the rwalk behaviour for difficult cases.

segasai:
This is a small but useful improvement. 
When doing random walk sampling the walking consists of two parts.
* Generating a point in a random ball/ellipse around the current point.
* Testing if it is accepted based on logl criteria

These steps are repeated 'walks' times. 
Sometimes even the first step generation of a point can fail, because say the ellipsoid is longer than the cube and a generated point is outside. Now I do not consider that as a 'walk'. I only consider walk attempts that are within the cube and that get a function evaluation. 
That should make things more stable and reduce situations when the none of walking steps are accepted (i.e. #354 )

### DIFF
--- a/py/dynesty/sampling.py
+++ b/py/dynesty/sampling.py
@@ -219,11 +219,10 @@ def generic_random_walk(u, loglstar, axes, scale, prior_transform,
     ncall = 0
     # Total number of Likelihood calls (proposals evaluated)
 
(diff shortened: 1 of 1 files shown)


segasai:
That was not a good idea. that still breaks the detailed balance. Closing.

PR #390, opened 2022-09-07
Results interface

segasai:
This is first part of the PR that I'd like to push in before the next release that allows to get equal weighted samples directly from the results object. I.e.

``` ns = dynesty.NestedSampler()'
res=ns.run_nested()
samps = res.samples_equal()
```
without having to do the manual calculation of weigths etc

For this PR I've had to move a lot of code from results.py to utils.py to avoid cyclic dependency issues. 
That's not great, but I didn't have a better way. 

Docs are still TBD, and feedback is welcome.


segasai:
I'll aim to merge this tomorrow barring any last minute objections.

joshspeagle:
None from me!

segasai:
Thanks. Merging..

Issue #368, opened 2022-04-15
Saving derived quantities during sampling


segasai:
This is the issue to discuss the way of saving the quantities derived during the likelihood function evaluations.
Sometimes the likelihood function does some heavy calculations and compute some derived quantities that are useful to look at after sampling. Currently the only way to look at those is to rerun the likelihood function evaluations on posterior samples. 

I think it is worth having the interface for that. (unless it's already possible now somehow)

My first guess is we can store the derived quantities in RunRecord https://github.com/joshspeagle/dynesty/blob/a0acba285a1ecabc4af380c43d286112fc74b6ef/py/dynesty/utils.py#L163 
The question is what is the best way to return those derived quatities, without having to write a lot of code in many places and making incompatible changes. 
We can possibly allow likelihood function return either a value or a tuple (logl, derived_quantities_vector) and hide the code inside this 
https://github.com/joshspeagle/dynesty/blob/eff15474063920c8e32bc6357e047515f5986cb2/py/dynesty/utils.py#L49 logl wrapper

tilmantroester:
I second that this would be a nice feature to have. At some point in the past this was considered too complicated to implement (https://github.com/joshspeagle/dynesty/issues/98) but I wonder if that has changed in the last ~3 years.

segasai:
I have started a test branch for possibly way of implementing this change 
https://github.com/joshspeagle/dynesty/tree/extra_info
that will still need a lot of work and it may still not be the right approach, but we'll see

The idea is to basically encapsulate the output of the loglikelihood function inside a special object that can store logl and auxiliary info and still support comparisons etc. This way in theory the changes to other bits of code should be small enough.

segasai:
This has been merged now

kcroker:
Thank you for adding blob support!  I'm running v2.1.2 and, while I can flag to store blobs, I am unable to recover them.  I just get an array of the same length as the samples, but containing None.  Any idea what might be going on here?

PR #395, opened 2022-09-21
Allow saving additional information returned by the likelihood function

segasai:
Starting implementation of #368 
The function can now return an additional blob
The tests should pass, but the additional info is not saved yet.
early feedback welcome


segasai:
This is now a workign version that allows the likelihood to return an arbitrary 1d numpy array (optionally) if blob=True option is specified in the sampler.  Those are stored in results.blob attribute
the example is in the test/test_blob.py
the docs are absent
It is unclear whether the name for the 'blob' is the best one,
the tests need to be expanded to include pools and resuming.

segasai:
I think this is ready for review. 
It'd be great to have a second look on this.
After excluding docs/tests the patch is surprisingly small and not too difficult. 
The performance effects are probably non-zero given the overhead of objects etc (at least the test suite on my machine seemed to slow down by ~ 2.5%, but that's probably small enough to not worry about it)
Also I am thinking of making the next version 2.0 given this change and the checkpointing..

PR #401, opened 2022-10-05
Resolve some issues when using custom sampler

ColmTalbot:
I ran into some issues trying to use a customised `Sampler` subclass and with a different proposal distribution.

The changes in this PR allow me to fairly simple define the new class and function to modify the analysis.

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -188,7 +188,11 @@ def _get_update_interval_ratio(update_interval, sample, bound, ndim, nlive,
         elif sample == 'hslice':
             update_interval_frac = 25.0 * slices
         else:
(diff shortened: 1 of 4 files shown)


segasai:
Thanks for adding a new test. that was previously lacking, because I never used this functionality
One quick comment is that I think it would be better to make the test not part of test_egg because here IMO we don't particularly care about the complex function, we just care that the code can run (i.e. a single gaussian like in test_gau would do).. Also I'm not sure the code needs to be part of the itertools.product over bounds too (i.e. custom+rslice and multi+sample_rwalk should do). 
The reason for these comments is that the test execution time is a concern, so one needs to make tests as short as possible.

segasai:
thank for the PR.
I've merged the edited version of it in #402

PR #428, opened 2023-03-05
Fixes to bounding updates

### LINKED ISSUE #427: Bounding fixes
Following up on #425 

I've stumbled upon this 
https://github.com/joshspeagle/dynesty/blob/2d49cf9104e417a44ada7f2f1d013cbfee0acf01/py/dynesty/nestedsamplers.py#L498

and similar code for MultiEllipsoid. Basically the default initialisation for these bounds is zero centre and unit covariance matrix which is obviously incorrect for an ellipsoid covering the whole cube. This is pretty inconsequential for slice/walk samplers, but this has many issues in particular for uniform sampler.

* That will sample a significant fraction of the volume outside the cube
* The sampling will be missing a significant volume and will be biased towards zero, at least till the first update of the boundary.

The correct initialisation should be centre at [0.5,...0.5] and covar of diag([ndim/4,...,ndim/4])

While I have a patch ready #428, I'm creating this issue for traceability (and feedback)

Comments ?

segasai:
Fix #427

When the ellipsoid is being initialized as a bound, it's centred at zero and has unit variance, while it should be centred at 0.5 and have cov of ndim/4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 - Change the way initialization of live-points happens. Now if after 1000 iterations we managed to get at least a single point, the code will continue sampling after putting a warning. Previously the error was raised. (by @segasai )
 - Refactor the bound update code which will lead to more consistent boundary updates (by @segasai )
+- If incorrect keys to first_update are provided an exception will be raised
 ### Fixed
 - Fix some pathological cases when uniform distribution is sampled with a very low logl.
 - Fix a problem when a very small nlive is used leading to the error (#424 , reported by @frgsimpson)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased]
 ### Added
 ### Changed
-- Change the way initialization of live-points happens. Now if after 1000 iterations we managed to get at least a single point, the code will continue sampling after putting a warning. Previously the error was raised.
+- Change the way initialization of live-points happens. Now if after 1000 iterations we managed to get at least a single point, the code will continue sampling after putting a warning. Previously the error was raised. (by @segasai )
+- Refactor the bound update code which will lead to more consistent boundary updates (by @segasai )
 ### Fixed
 - Fix some pathological cases when uniform distribution is sampled with a very low logl.
 - Fix a problem when a very small nlive is used leading to the error (#424 , reported by @frgsimpson)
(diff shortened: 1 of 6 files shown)


segasai:
This is ready to get merged (it'd be great to get a review though).

The reason I'm not merging this yet is because when running test_resume.py one can see sometimes a slowdown when resuming the run (revealed when looking at #432 ). I think this means that some necessary information about bound update status is not fully propagated when saving/resuming. 

Snippet from the interrupted run 
```
475it [00:04, 81.38it/s, batch: 0 | bound: 0 | nc: 16 | ncall: 1197 | eff(%): 31.730 | loglstar:   -inf < -13.512 <    inf | logz: -17.612 +/-  0.102 | dlogz: 16.065 >  0.010]terminating
resuming
634it [00:00, 6329.26it/s, batch: 0 | bound: 0 | nc: 8 | ncall: 2375 | eff(%): 26.695 | loglstar:   -inf < -8.134 <    inf | logz: -12.384 +/-  0.102 | dlogz: 11267it [00:00, 1236.87it/s, batch: 0 | bound: 1 | nc: 2 | ncall: 11294 | eff(%): 11.218 | loglstar:   -inf < -0.917 <    inf | logz: -5.107 +/-  0.104 | dlogz: 1
```
as opposed to the non-interrupted run

```
468it [00:00, 4677.96it/s, batch: 0 | bound: 0 | nc: 3 | ncall: 1187 | eff(%): 31.473 | loglstar:   -inf < -13.867 <    inf | logz: -17.886 +/-  0.102 | dlogz: 936it [00:00, 2893.89it/s, batch: 0 | bound: 0 | nc: 9 | ncall: 6559 | eff(%): 13.646 | loglstar:   -inf < -2.738 <    inf | logz: -6.959 +/-  0.104 | dlogz:  31263it [00:00, 1620.68it/s, batch: 0 | bound: 1 | nc: 4 | ncall: 10713 | eff(%): 11.468 | loglstar:   -inf < -0.904 <    inf | logz: -5.014 +/-  0.103 | dlogz: 
1483it [00:00, 1354.41it/s, batch: 0 | bound: 2 | nc: 3 | ncall: 11127 | eff(%): 12.978 | loglstar:   -inf < -0.420 <    inf | logz: -4.539 +/-  0.103 | dlogz: 1
```
One cam see the difference in ncalls

and the interrupted/resumed run takes ~ 2 min as opposed to 10 sec.
The issue seems to occur with the pool only.

Also it's not quite clear if this is really a new bug introduced with #428 or not, since the previous bounding code was somewhat obscure to me.

segasai:
I've merged this.
The issue with the slow down of test/resume is not connected to the patch, but instead just a result 
of starting the batch with logl value below the logl value of the first bound.

joshspeagle:
Thanks so much! (Sorry for being slow -- the end of the semester has been a bit crazy!)

segasai:
No worries. 
I've just created one additional bounding issue that may need addressing in the future.

Issue #438, opened 2023-05-03
Dynesty dynamic sample resuming from checkpoint does not use pool


segasai:
** Dynesty version **
2.0.0-2.1.1 

**Describe the bug**
The resuming of dynamic nested sampler when using a pool does not properly restore a pool.

This can be seen while running this script. (first to create one checkpoint file, and then restore from it).
After restoring the pool is not used.

```python
import numpy as np
import multiprocessing as mp
import dynesty
import sys


def func(x):
    ret = -(x**2).sum()
    for i in range(1000000):
        i**2
    return ret


def priort(x):
    return x * 20 - 10


def main(fname, nthreads=35, dynamic=True):
    ndim = 10
    if nthreads == 1:
        pool = None
    else:
        pool = mp.Pool(36)
    CL = get_class(dynamic)
    dns = CL(func, priort, ndim, pool=pool, queue_size=100)
    dns.run_nested(checkpoint_file=fname)


def get_class(dynamic):
    if dynamic:
        CL = dynesty.DynamicNestedSampler
    else:
        CL = dynesty.NestedSampler
    return CL


def resume(fname, nthreads=35, dynamic=True):
    if nthreads == 1:
        pool = None
    else:
        pool = mp.Pool(36)
    print(pool, pool.map)
    CL = get_class(dynamic)
    dns = CL.restore(fname, pool=pool)
    dns.run_nested(resume=True)


if __name__ == '__main__':
    dynamic = True
    fname = 'xx.pkl'
    if sys.argv[1] == '0':
        main(fname, dynamic=dynamic)
    else:
        resume(fname, dynamic=dynamic)
```

I'll submit the PR in a second. This issue is just for tracing/discovery purposes.

segasai:
#439 fixed this issue. Closing

PR #439, opened 2023-05-03
fix the resuming of dynamic nested sampler runs with the pool

segasai:
Addresses #438 

What remains to be done here is some kind of test that would prevent this bug from happening.
I.e. the test that would verify that after restoring with the pool, it's actually used. 

If I won't have an idea for the test, I'll be merging without.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 ### Fixed
 - Fix the restoration of the dynamic sampler from the checkpoint with the pool. Previously after restoring the sampler, the pool was not used. (#438 ; by @segasai)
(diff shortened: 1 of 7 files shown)


segasai:
Now with the last commit I included the test that properly verifies that the pool is used after the restoration from the checkpoint.  That helped me to discover that the first patch only fixed the case if the restoration from the checkpoint happens during the initial run.
I think I'll merge this in a couple of days and probably make a new release after that.

joshspeagle:
Good catch on this and thanks for quickly addressing the bug!

segasai:
Okay, this thread has two more fixes of issues that I uncovered when running the tests with different seeds. 
* Previously if the batch run took shorter than the checkpoint_every nothing was saved. 
* When checkpointing, previously one last point before the checkpoint could have been lost leading to the differences in outputs of interrupted+restarted run vs pristine run.

Issue #440, opened 2023-05-12
issue with weights and posteriors comparing v0.9.5.3 and v2.1.1


mmfausnaugh:
** Dynesty version **
v2.1.1, `pip install dynesty`

**Describe the bug**
This is an extension of  issue #381 .  Basically, I am seeing inconsistent results in the weights and posteriors on an old computer using v0.9.5.3 and a new Apple M1 chip with v2.2.1.

I updated example code at https://github.com/mmfausnaugh/SN_analysis_debug/tree/main
To run the test, you would clone the directory and do python analyze_SN_curved_PL_dnest.py

 v0.9.5.3 behaves as I expect, and accpetence fractions are about 20%.  With v2.1.1., I get acceptance fractions less than 1%, and the posteriors for one of my tests looks very strange.  

**Setup**
See example code in repository: https://github.com/mmfausnaugh/SN_analysis_debug/tree/main

** Dynesty output **
The example code with save the output from dynesty as a .npz.  The biggest issue is that with 2.1.1., the weights are set to < 1.e-20 for one of my tests.

**Bug**

The code runs to completion, but the posterior I get for one object causes problems.  I can't actually plot it with dynesty functions---you can reproduce my issues running `python plot_dynesty_chains.py test_2022exc_dynesty_results.npz' after running the above example code.  Here is the output that is failing in dynesty:

```
/Users/faus/miniconda3/envs/py310/lib/python3.10/site-packages/dynesty/plotting.py:1368: UserWarning: Attempting to set identical low and high ylims makes transformation singular; automatically expanding.
  ax.set_ylim([0., max(n) * 1.05])
/Users/faus/miniconda3/envs/py310/lib/python3.10/site-packages/dynesty/plotting.py:2310: RuntimeWarning: invalid value encountered in true_divide
  sm /= sm[-1]
WARNING:root:No points at all in the plotted region
Traceback (most recent call last):
  File "/Users/faus/SN_analysis/SN_analysis_debug/plot_dynesty_chains.py", line 46, in <module>
    fig,axes = dyplot.cornerplot(results,
  File "/Users/faus/miniconda3/envs/py310/lib/python3.10/site-packages/dynesty/plotting.py", line 1451, in cornerplot
    _hist2d(y,
  File "/Users/faus/miniconda3/envs/py310/lib/python3.10/site-packages/dynesty/plotting.py", line 2364, in _hist2d
    ax.contourf(X2,
[...]

PR #443, opened 2023-05-18
Boundary updates

segasai:
This is a tentative fix to the BIC criterion for ellipsoid splitting introduced in #286 
There a factor of two was skipped in the calculation making the splitting of ellipsoids harder.
Apparently that leads to significant slowness observed in #440

I'm a bit torn by this issue and whether it's right change. One wouldn't want the results to be that sensitive to that factor of two here and I'm concerned if this will break some other cases... 

One thing that this change trips is the test_ellipsoids test case https://github.com/joshspeagle/dynesty/blob/d9164354421f4a010ee2bc232cb4eda2f4f90a88/tests/test_ellipsoid.py#L265 where I check if the number of clusters in 6 x 6 x  6 x 6 grid  in 4d is within 10% of 6^4. 
I don't know if it's a serious concern or not... 

Thoughts ? 

(sorry for creating a 2nd PR on this, the first one had an incorrect branch by mistake)


segasai:
Here's analysis of the performance number of clusters recovered

image

for the old version vs new version of the code for different number of clusters N=1,2,5,10,100,1000

Here's the code.  I generate points on a grid, randomly scale it and randomly rotate it.
I'd say it looks like an improvement.

```python
import itertools
import numpy as np
import dynesty.bounding as db
import math
import scipy.linalg
import multiprocessing as mp


def random_mat(rstate, ndim):
    # random rotation matrix
    x = np.random.normal(size=(ndim, ndim))
    cov = np.cov(x)
    mat = scipy.linalg.svd(cov)[0]
    return mat


def test_number_clusters(ndim, ncens, nper_dim=10, sig=0.01, seed=1):
    # check we can recover close to a correct and large number of clusters

    npt = ncens * nper_dim * ndim
    rstate = np.random.default_rng(seed)
    # create stuff on the grid
    nxcens = int(math.ceil(ncens**(1. / ndim)))
    cens = np.zeros((ncens, ndim))
    curiter = itertools.product(*[range(nxcens)] * ndim)
    for i in range(ncens):
        cens[i] = next(curiter)
    xs = sig * rstate.normal(size=(npt, ndim)) + cens[np.arange(npt) %
                                                      len(cens), :]
    # scale dimensions
    min_scale = 1e-3
    scales = 10**rstate.uniform(np.log10(min_scale), 0, size=ndim)
    xs = xs * scales[None, :]
    # random rotation
    mat = random_mat(rstate, ndim)
    xs = xs @ mat
    E = db.bounding_ellipsoids(xs)

    return ncens, len(E.ells)


def doall(ntest=10000, nthreads=36):
    rstate = np.random.default_rng(2)
    ndims = rstate.integers(2, 20 + 1, size=ntest)
    ncens = np.array([1, 2, 5, 10, 100, 1000])
    xcens = rstate.integers(0, len(ncens), size=ntest)
    ncens = ncens[xcens]
    nper_dims = rstate.integers(2, 10 + 1, size=ntest)
    res = []
    with mp.Pool(nthreads) as poo:
        for i in range(ntest):
            res.append(
                poo.apply_async(test_number_clusters,
                                (ndims[i], ncens[i], nper_dims[i], 0.01, i)))
        for i in range(ntest):
            if i % 1000 == 0:
                print(i)
            res[i] = res[i].get()
    return res
```

segasai:
After some more thinking/testing. 
I have

* scrapped the BIC change as it was leading to failures on the grid tests (and #440 is not good enough justification for the change)
* After some reading on kmeans  I decided that it is a good idea to rescale the points to the same variance along each axis. I didn't change the initialization of kmeans, although there is some interesting reading here  ( https://doi.org/10.1016/j.patcog.2019.04.014 ) 
* I have added one warning that complains if the bootstrapping factor is too large (i.e. if inflation is > 100 in volume).  Currently that warning is repeated each time the error is encountered. 
* I have added a warning if the sampling is stopped due to maxcall or maxiter -- now the warning is printed that the posterior may not be properly sampled if the stopping happened if dlogz was large enough) -- that would prevent the #440 complaint .

@joshspeagle  comments ?

segasai:
Merged this as this have been have been sitting for too long

Issue #456, opened 2023-10-26
Using dynesty with npdim


MichaelDAlbrow:
** Dynesty version **
dynesty 2.1.2 installed with pip

**Describe the bug**

This may be a bug or a fundamental misunderstanding on my part about the dimensions of the input and output of prior_transform. Please let me know if this is the case, and my advance apologies for raising this as a bug. 
(Michael Albrow)

I am trying to use dynesty with npdim < ndim. My prior transform accepts npdim parameters and returns ndim parameters.

_initialize_live_points in dynamicsampler.py has a call:

cur_live_v = M(prior_transform, np.asarray(cur_live_u))

or 

cur_live_v = map(prior_transform, np.asarray(cur_live_u))

followed a little later by

live_v[ngoods:ngoods + nextra] = cur_live_v[cur_ind]

Earlier in the function, live_v is initialised with length npdim, but cur_live_v (being the output of the prior_transform)
is of length ndim.

** Dynesty output **
Traceback (most recent call last):
  File "/home/users/mda45/Microlensing/Galacticmodel_compress_10Nov21/ob170108.py", line 210, in <module>
    sampler = dynesty.NestedSampler(loglike, prior_transform_total_samples, 
  File "/home/users/mda45/miniconda3/lib/python3.10/site-packages/dynesty/dynesty.py", line 677, in __new__
    live_points, logvol_init, init_ncalls = _initialize_live_points(
  File "/home/users/mda45/miniconda3/lib/python3.10/site-packages/dynesty/dynamicsampler.py", line 458, in _initialize_live_p
    live_v[ngoods:ngoods + nextra] = cur_live_v[cur_ind]
ValueError: could not broadcast input array from shape (227,10) into shape (227,4)

segasai:
Thanks for the report. 

I can say I never quite understood what npdim is supposed to do (what is the use-case) and there is not a single test written for that behaviour. So it is possible it got broken at some point.

Did this work in previous dynesty versions ? 

This description " Number of parameters accepted by `prior_transform`. This might
            differ from `ndim` in the case where a parameter of loglikelihood
            is dependent upon multiple independently distributed parameters,
            some of which may
            be nuisance parameters." from dynesty.py doesn't really clarify much. 

Can you comment on this @joshspeagle  ?

MichaelDAlbrow:
When I open dynamicsampler.py in github, and click on "blame", it seems that this code section may have been updated 10 months ago (line 422).

[...]

PR #457, opened 2023-10-27
get rid of npdim

segasai:
This gets rid of npdim and produces a couple of deprecation warnings.

A couple of decisions I had to make -- I made sure to rename all the npdim= keywords in various classes (other than the top ones like NestedSampler and DynamicNestedSampler) to ndim=. The reason is that I prefer minimize the number of concepts/standard variable names throughout the code.
The consequence of that is that when we release this i don't quite know if restoring checkpoints from previous version will be possible. I don't think it is a big deal, but not everyone may agree.

This is followup from #456

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 [Unreleased]
 ### Added
+
 ### Changed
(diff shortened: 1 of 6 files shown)


segasai:
I checked that the checkpoints from previous version will not be restored with this change. The error message https://github.com/joshspeagle/dynesty/blob/f59d963fad80301e5d28bc5a6e3718c467c1bf94/py/dynesty/utils.py#L2272  
 however is clear enough, so I don't consider that an issue.
 I'll merge this soon.

PR #465, opened 2023-12-11
add missing h in docs front page

eteq:
I noticed this while doing #464 . It's about as trivial a PR as can exist, but I can't un-see it now! :wink:


segasai:
Thanks for spotting that!

PR #466, opened 2023-12-11
Pool args fix

segasai:
Address #464

### DIFF
--- a/py/dynesty/pool.py
+++ b/py/dynesty/pool.py
@@ -107,6 +107,9 @@ class Pool:
 
     as this way they will not need to be pickled and unpickled every function
     call.
+    
+    Note though that if you specify logl_args, and ptform_args when  creating
+    the Pool *AND* in the sampler those will be concatenated
     """
 
     def __init__(self,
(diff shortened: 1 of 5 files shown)


py/dynesty/pool.py
```diff
@@ -93,6 +96,17 @@ class Pool:
...
+    If your likelihood function takes additional arguments, it is better to
+    pass them when creating the pool, rather then to nested sampler::
+
+        with dynesty.pool.Pool(16, loglike, prior_transform, 
+                                            logl_args=(...) ) as pool:
+            dns = DynamicNestedSampler(pool.loglike, pool.prior_transform, ndim,
+                                     pool=pool)
+
+    as this way they will not need to be pickled and unpickled every function
+    call.
```
eteq, line 109: I think it'd be good to also include this info here even though it isn't strict the pool itself so much as the sampler initialization.
```suggestion
    call.
    
    Note though that if you specify logl_args, and ptform_args when  creating
    the Pool *AND* in the sampler those will be concatenated
```

eteq (approved):
Looks good to me as a better fix for #464 !

One minor doc suggestion though, just to make sure the user catches that the concatenation behavior happens.

segasai:
Thanks for taking a look. I accepted your doc edit, but I didn't catch what you meant by this (or whether it was what exactly about your doc edit)

> One minor doc suggestion though, just to make sure the user catches that the concatenation behavior happens.

I think the concatenation is mentioned now both in pool.py, and docs (and I don't think the runtime warning is needed here)

PR #467, opened 2024-01-16
Convergence testing with distance travelled

ColmTalbot:
### Reference issue

Motivated by discussion #365 and #366

Hi @segasai, first of all I'm sorry for my extended absence on #365 and #366 I went down a deep rabbit hole related to the discussion in #365 and that combined with other work/life commitments didn't leave much time for implementing this. I think I've got to the bottom of this and if you're still willing I'd like to get back into this.

I've attached a draft describing a new test that uses the distance travelled in the MCMC as a diagnostic that is implemented in this branch.

I think this test and toy model problems we've discussed before are an excellent pair to put together a set of tests for potential changes to the sampling/MCMC methods.

[nested_sampling_distance_test.pdf](https://github.com/joshspeagle/dynesty/files/13951394/Bilby_dynesty_proposals.pdf)

I've included a new plot that shows the likelihood/distance insertion index

image

I've also been experimenting with plots that will show evolution of the KL divergence estimate, e.g., 
image

### DIFF
--- a/py/dynesty/dynamicsampler.py
+++ b/py/dynesty/dynamicsampler.py
@@ -605,6 +605,7 @@ def _configure_batch_sampler(main_sampler,
     saved_scale = np.array(main_sampler.saved_run['scale'])
     saved_blobs = np.array(main_sampler.saved_run['blob'])
     saved_distances_indices = np.array(main_sampler.saved_run["distance_insertion_index"])
+    saved_log_distance_ratios = np.array(main_sampler.saved_run["log_distance_ratio"])
     saved_likelihood_indices = np.array(main_sampler.saved_run["likelihood_insertion_index"])
     first_points = []
 
@@ -724,6 +725,7 @@ def _configure_batch_sampler(main_sampler,
 
         live_scale = saved_scale[subset0[0]]
         live_distance_index = saved_distances_indices[subset0[0]]
+        live_log_distance_ratio = saved_log_distance_ratios[subset0[0]]
         live_likelihood_index = saved_likelihood_indices[subset0[0]]
         # set the scale based on the lowest point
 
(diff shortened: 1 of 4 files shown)


segasai:
Hi @ColmTalbot , 

Thanks for the updated patches. 
I will try to take a look at them in next couple of weeks ( a bit busy with teaching now)


     Sergey

segasai:
Hi, @ColmTalbot 

Thanks for the patches. I looked at them and I am certainly a fan of the logl insertion index statistic in particular. It would be great to have that convergence test. 

I have however a suggestion of somewhat different implementation to the one you are proposing.The current implementation is in my opinion somewhat invasive, as it adds a bunch of keywords/variables specific to your convergence statistics that needs to be stored during sampling. 
However what I have in mind is to make sure we store enough general information during the sampling, so that the convergence statistics (like the ones you are proposing, or some other ones) can be calculated after the run.

In my understanding, if we, for example, store the iteration at which each point is added, i.e. 
On top of live_u, live_v, live_logl arrays in Sampler() we create the a live_additer or something array, that can track when each point was added, and that can the be propagated further when we take out dead points, and that will be saved in saved_run.RunRecord() 
In the case of static run, I think it is trivial to use that info to compute the logl order stastistic (and likely distance statistic) after the fact. In the case of dynamic run things are more complex, but I think doable (maybe more information needs to be stored about the run, but so be it).

Basically my thinking is that I'd prefer these convergence statistics to be computed after the run (outside the sampler) and we just need to ensure the relevant information is saved, so we can properly retrace the steps of the sampler. 

Thanks 
      S

segasai:
I agree that it can be useful to know some convergence statistics during the run. I would put that lower in priority for the moment. 
Regarding the distance travelled, I'm inclined to store in the RunRecord the proposed starting point for MCMC-based (rather than uniform) samplers. That again would enable computing the distance statistic in postprocessing.

Issue #391, opened 2022-09-09
Generic sampler interface


segasai:
While consider PR's like #365, #366
I come to think that the best architecture is to have 
a class interface that is responsible provide new samples. I.e. 
RSlice,  RWalk and so forth. The goal of this would be to simplify adding new samplers, like ensemble sampling.
Also in theory that would enable passing samplers directly to NestedSampler() at construction, i.e.
```NestedSampler(sampler=RSlice(slices=4)) ```

This encapsulation potentially also will do the tuning inside those classes. Currently the tuning of proposals is sort of done at high level like here
https://github.com/joshspeagle/dynesty/blob/2236388ac60e9cf9a85dc57636c1b5a9c322c92a/py/dynesty/nestedsamplers.py#L140 

Another possible benefit is to enable combining samplers, like 

```python
NestedSampler(sampler=[RSlice(slices=4),Rwalk(waks=5)])
```

The key questions is what exactly the interface of those samplers should be. 
I.e. for example, it is not clear to me that is useful to keep the separation of 'propose live-point' vs 'evolve live-point' methods or whether each method should just do the job of 
* You are provided with a list of points subject to L>L*
* You are optionally provided with a list of ellipsoids 
* You need to provide a new point subject to L>L* 

Also, it is not quite clear how much state should be preserved in those samplers, and whether it can be a problem for multi-core runs, if we have to send/pickle those samplers, as well as what should be done with those sampler states on different cpus.

ATM this is not a very concrete proposal, more of an RFC. 
In the simplest form, the proposal would just bundle sample_rwalk(), sample_rslice() into 
sample() methods of a class, but that's an easy part.

segasai:
Thinking more about the design of this, the main complexity is the parallel sampling. 
Ideally one want to have a sampler object like RWalk() keep track of tuning and do the sampling, but then 
we'd have to send this big object to the pool workers when sampling (that's potentially expensive), but even a more difficult problem is how to combine the results after, i.e make sure we update the scale factors.

segasai:
I think I have a sketch of a possible new interface.

Again the point is to have samplers as a distinct class that is quite independent from the rest and could be easily replaced. I.e. #366

The idea is to have a two steps in the sampler. One is for running on the main thread/preparing, and the other ones doing the actual sampling. 

The actual class will do the internal book-keeping, tuning (i.e. what's currently done by update_rwalk  etc)
The actual sampling will be a staticmethod to avoid sending the whole class over

```python
[...]

PR #471, opened 2024-02-14
Parallel uniform sampling + New sampler interface

segasai:
This is a preliminary version of  patch that enables parallel uniform sampling.
The idea is to get away from this propose/evolve for uniform distributions where evolve doesn't really do anything other than evaluate logl on a single pt, but instead make propose a no-op for uniform, and evolve() actually do the sampling inside the boundary till a satisfactory pt is found. 

While implementing this I realized that after some changes all the different samplers in nestedsamplers.py may not need to be there as they do something very similar (but that's tbd). 

ATM some tests fail, but I think that's fixable.

Also some class members had to be renamed. I.e. .bound attribute is really misleading in the classes as it was storing the list of all the historic bounds, so I renamed that to bound_list.


joshspeagle:
!!!! This is an **excellent** proposed change! Very interested to see how the other internal refactoring might also work.

segasai:
Thanks @joshspeagle 
Just out of curiosity i've actually refactored nestedsamplers. It was surprisingly easy to get rid of those classes there which removed few hundred lines (of codes + comments).
I am not 100%, but 95% certain that's a good change just because it's easier to understand it now (and there was so little different between those samplers)

segasai:
The demonstration of this patch. The following uniform sampling of Eggbox with single bound 
takes 36 minutes:
9528it [36:29,  4.35it/s, +1000 | bound: 8600 | nc: 1 | ncall: 12936194 | eff(%):  0.081 | loglstar:   -inf < 243.000 <    inf | logz: 235.824 +/-  0.079 | dlogz:  0.000 >  0.100]
With the patch it takes 3 minutes.

9490it [03:16, 48.25it/s, +1000 | bound: 999 | nc: 1 | ncall: 12943702 | eff(%):  0.081 | loglstar:   -inf < 243.000 <    inf | logz: 235.863 +/-  0.078 | dlogz:  0.000 >  0.100]


```python
import numpy as np
import dynesty
import multiprocessing as mp

nlive = 1000


def loglike_egg(x):
    logl = ((2 + np.cos(x[0] / 2) * np.cos(x[1] / 2))**5)
    return logl


def prior_transform_egg(x):
    return x * 10 * np.pi


LOGZ_TRUTH = 235.856


def test_bounds():
    bound, sample = 'single', 'unif'
    # stress test various boundaries
    ndim = 2
    rstate = np.random.default_rng(444)
    with mp.Pool(4) as poo:
        sampler = dynesty.NestedSampler(loglike_egg,
                                        prior_transform_egg,
                                        ndim,
                                        nlive=nlive,
                                        bound=bound,
                                        sample=sample,
                                        rstate=rstate,
                                        pool=poo,
                                        queue_size=4)
        sampler.run_nested(dlogz=0.1, print_progress=True)
        assert (abs(LOGZ_TRUTH - sampler.results.logz[-1])
                < 5. * sampler.results.logzerr[-1])


if __name__ == '__main__':
    test_bounds()
```

segasai:
It would be great to have some review of this patch @joshspeagle if you have time, before the patch gets bigger. 

My current thinking is the patch is mostly positive with a few negatives listed below:
*  Previously I had a test of custom sampler like that  _SAMPLERS["custom"] = MultiEllipsoidSampler. I had to disable this, but I don't actually know if anyone in the wild is using this functionality.
* The new logic in propose_live has too many ifs() for my liking because it has to deal with combinations of different bounds/samplers. I think this needs to be cleaned up.
* kwargs is used to communicate the bound and a few other variables to sample_bound_unif. Maybe that needs to be changed and we need to expand SampleArgument https://github.com/joshspeagle/dynesty/blob/3f100253dd53fd302ee84f7f25f010b3bf70ad29/py/dynesty/sampling.py#L24 rather than abuse kwargs
* Renaming of a few attributes (there were .method/.sampling attributes, now there is only .sampling). The .bound is renamed to .bound_list.
* Some interface change to bounding classes, so they can be used interchangeably. Maybe having an BoundingInterface class is a good idea...
* I don't have an actual example, but I could imagine the case where the new implementation is slower. Maybe the case of very high acceptance rate, very fast logl function, and very large bound, in this case the overhead from pickling the bound and sending it to the pool worker may be significant. Probably this is not a big worry, since if acceptance fraction is high, things are easy no matter what.

Among things I'd like to do, but I don't want to do in this patch: I think with this patch it's becoming easier to do #391. We already now can pack update_slice or update_rwalk in SliceSampler RWalkSampler objects and make sure that things like history https://github.com/joshspeagle/dynesty/blob/3f100253dd53fd302ee84f7f25f010b3bf70ad29/py/dynesty/nestedsamplers.py#L145 is stored inside those objects. But I'd prefer to do that separately, as this requires more thought.

segasai:
Thanks for the link @ColmTalbot. I think those interfaces will certainly break with this patch, which is I think unfortunate, but I think if we combine this change with the new sampler interface ( #391), hopefully it'll be worthwhile.
I did particularly want to implement the ensemble (a-la emcee) proposal in any new sampler interface.

What I need to implement the new way of sampling is to formalize the exact interface for those 'inner samplers' (rwalk/slice/uniform)

It is clear each sampler needs the sample()  interface as defined in sampling.py, 
They also need update() interface as currently written in update_slice() etc.
But I think my understanding is some samplers also need extra arguments passed to evolve_point,
so therefore each sampler also needs some sort of generate_args(Sampler). In the simplest form it would just set the kwargs to whatever you want.

segasai:
After quite some time, I finally was able to finish this and also #391 
Now the bounding, and the internal sampler are (almost) completely separate interfaces. 

The samplers also themselves keep track of necessary tuning.
In theory this enables trivially implement new samplers and bounding.

In the process I've also got rid of hslice, as it was failing the tests for unclear reason and the code was never properly validated.

Since the code passes all the tests, I'm quite confident the change is fine. What I will further work on is on the interface and how samplers arguments are passed around. 

The changes here have potential to break some codes, so this will be either 2.2 version, or 3.0.  (

If I have time over next month, I'll probably implement the ensemble slice sampler as a test of the new interface. 
I also may clean up other small things given that this will be breaking release.

PR #478, opened 2024-06-21
Replace x.ptp() with np.ptp(x)

joezuntz:
Numpy 2.0 removes the array method `.ptp()` in favour of the function `np.ptp`. This replaces the handful of calls in the code.


segasai:
Thanks @joezuntz ! I wasn't aware of the .ptp deprecation. I'll merge this after tests run

segasai:
The new version of dynesty on pypi contains this fix

Issue #492, opened 2024-12-23
Setup.py uses a removed `pypandoc.convert` method


xlucn:
** Dynesty version **
2.1.4. From Arch Linux's AUR package, which builds and installs with source files.

**Describe the bug**
During build (`python -m build --wheel --no-isolation`), I run into an error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.13/site-packages/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
    main()
    ~~~~^^
  File "/usr/lib/python3.13/site-packages/pyproject_hooks/_in_process/_in_process.py", line 373, in main
    json_out["return_val"] = hook(**hook_input["kwargs"])
                             ~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/pyproject_hooks/_in_process/_in_process.py", line 143, in get_requires_for_build_wheel
    return hook(config_settings)
  File "/usr/lib/python3.13/site-packages/setuptools/build_meta.py", line 332, in get_requires_for_build_wheel
    return self._get_build_requires(config_settings, requirements=[])
           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/setuptools/build_meta.py", line 302, in _get_build_requires
    self.run_setup()
    ~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/setuptools/build_meta.py", line 516, in run_setup
    super().run_setup(setup_script=setup_script)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/setuptools/build_meta.py", line 318, in run_setup
    exec(code, locals())
    ~~~~^^^^^^^^^^^^^^^^
  File "<string>", line 24, in <module>
AttributeError: module 'pypandoc' has no attribute 'convert'
```
Which points to this line in code:
https://github.com/joshspeagle/dynesty/blob/55d93d00231d6eb8d8ddc534416ea7f1de8a2944/setup.py#L24

[...]

PR #493, opened 2024-12-23
use pypandoc.convert_text method, fixes a potential error during setup

xlucn:
### Reference issue

#492

`pypandoc.convert` was deprecated and removed in pypandoc. The usage can be replaced with `pypandoc.convert_text` with no other changes.


segasai:
Thanks for the report and fix. 
I have in the pipeline the migration to the pyproject.toml based installation, but for the time being this seems like correct fix.

PR #498, opened 2025-05-14
MAINT: avoid numpy deprecation error in empty intersection

### LINKED ISSUE #497: numpy deprecation issue
** Dynesty version **
v2.1.5 from pypi (also verified present on master)

**Describe the bug**

Checking whether a numpy array is empty by boolean comparison is deprecated. Fortunately the fix is well documented in the error message. I'll open a PR now.

```python
ndim = 2, periodic = [0], reflective = [1]

    def get_nonbounded(ndim, periodic, reflective):
        """
        Return a boolean mask for dimensions that are either
        periodic or reflective. It will be true for normal dimension
        and false for periodic/reflective
        """
        if periodic is not None and reflective is not None:
>           if np.intersect1d(periodic, reflective) != 0:
E           ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```

ColmTalbot:
### Reference issue

Closes https://github.com/joshspeagle/dynesty/issues/497

Switch to the numpy recommended method to check for an empty array.

cc @segasai @joshspeagle

### DIFF
--- a/py/dynesty/utils.py
+++ b/py/dynesty/utils.py
@@ -821,7 +821,7 @@ def get_nonbounded(ndim, periodic, reflective):
     and false for periodic/reflective
     """
     if periodic is not None and reflective is not None:
-        if np.intersect1d(periodic, reflective) != 0:
+        if np.intersect1d(periodic, reflective).size > 0:
             raise ValueError("You have specified a parameter as both "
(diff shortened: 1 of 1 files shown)


segasai:
Thanks for the patch! 

Merging

PR #507, opened 2026-06-13
fix bugs found in code audit

joshspeagle:
- fix crash in SupFriends.update with mc_integrate=True and store .ctrs
  consistently with RadFriends
- fix inverted plateau warning in unravel_run
- fix quantile() crash on single weighted sample and make it always
  return an array as documented
- fix broken code path in Sampler.update_bound_if_needed when
  bound_update_interval is None
- make run_nested(maxiter=N) perform at most N iterations
- validate slices >= 1 in slice samplers and stop silently replacing
  slices=0/walks=0 with defaults in the factory functions
- fix missing f-string in kld_error error message
- remove dead code (phantom 'nonperiodic' kwarg in slice samplers,
  duplicate rstate init in traceplot, dead update_interval branch in
  sample_initial)
- rename plot_thruth to plot_truth and fix incorrect docstring defaults
  in plotting and sampler modules

https://claude.ai/code/session_014g8GQqXDRpqCjXyAjJNJgj

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 [Unreleased]
 ### Added
+- The `slice` and `rslice` samplers now respect `periodic` and `reflective`
+  boundary conditions (previously they treated every dimension as a hard
+  unit-cube wall), improving sampling of posteriors whose modes lie on or near
+  a boundary.
 ### Changed
+- `sample='auto'` now honors an explicitly supplied `walks`/`slices` count
+  instead of silently using the default, and warns when a step-count option
+  does not apply to the chosen sampler.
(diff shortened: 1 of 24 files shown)


segasai:
Thanks. But I am against merging this as it is. This mixes whitespace/typo changes with modules structure refactoring with logical changes with unclear consequences. Some of the fixes would also require clear tests. Also from a quick look most of this does not look like actual pure bugs that give existing users incorrect results. 

Since some/most of the changes are probably good I would be happy to do review single individual focused PRs based on this, if there is a human behind each PR. otherwise I will keep this as is and then only when I have time I may cherry pick some of this.

joshspeagle:
Works for me. I was mostly trying to run some tests of behaviour on the repo with Opus/Fable as well as CoPilot with CI, so none of this is actually time-sensitive or all that important and I wasn't intending to merge the PR without external review. You can feel free to close this out whenever.

PR #509, opened 2026-07-20
feat(cornerplot): extend `title_fmt` to accept lists as well

mbsck:
### Reference issue

No reference issue exists.

Enable the user to pass a list for `title_fmt` to extend the possibility of different title formats (e.g. dimensions where `.2f` would show `0.00` could be set to `.2e`)

### Addressing the contribution policy
* No LLM was used in this pull-request
* I didn't see plotting tests and this change is very straightforward (do tell me if a tests is required anyways)

### DIFF
--- a/py/dynesty/plotting.py
+++ b/py/dynesty/plotting.py
@@ -1176,9 +1176,9 @@ def cornerplot(results,
         A list of fractional quantiles to use in the title. Default is
         `[0.025, 0.5, 0.975]` (median plus 95%/2-sigma credible interval).
 
-    title_fmt : str, optional
-        The format string for the quantiles provided in the title. Default is
-        `'.2f'`.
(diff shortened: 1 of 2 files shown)


segasai:
Thanks! It looks fine to me. But there is a bit of testing in test_plot.py. Please add something there. Basically it'd be nice to have both branches of the code covered. 
Also in the function docstring, you should probably say that if the list provided it is expected to match the ndim.

mbsck:
Thanks for the quick response! I force-pushed an extension of the `plot_test` for the `title_fmt` onto the same commit and it passed. I also adapted the docstring to mention the requirement of the matching dimensionality.

segasai:
Looks good. I'm merging.
Thanks for the contribution!

PR #511, opened 2026-08-04
Fix the treatment of the initial prior volume for likelihoods with -inf

segasai:
When the likelihood is -inf over part of the prior, the live-point initialization estimate fraction and records its logvol_init = -log(iattempt). The progressive integration during sampling uses this correctly, but code paths that reintegrate a finished run assumed the run started at unit prior volume:

- the final recomputation of the integrals at the end of the static
run_nested() 
- jitter_run(), resample_run() and unravel_run(), which rebuilt logvol as cumulative sums starting from zero
- merge_runs()/_merge_two(), which restarted the merged volume walk at logvol
= 0,
- the analogous recomputations in the dynamic sampler  (sample_initial baseline finalization and combine_runs).

In practice the issue is unlikely to affect users unless they are doing manual merge_runs(), and the significant fraction of the prior has logL=-inf

The issue was identified by Claude & Codex, and fix has been developed by me&Claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 [Unreleased]
 ### Added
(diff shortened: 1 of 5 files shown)


joshspeagle:
Wow, that is some edge case! Good find.

